PR #42, opened 2025-10-23
Add support for creating a C# addon to create-addon command

JesseCol:
* Added `create-addon --template=cs` command that creates a C# add-on and installs the node-api-dotnet package.
* Updated samples/electron to use a C# add-on as well.  The sample uses AoT and works both unpackaged and packaged.
* winsdk restore automatically updates package versions in Directory.packages.props to match winsdk.yaml.
* Also I updated the icons for the samples/electron app.

The command and output looks like this:

```
PS D:\work\w> npx winsdk node create-addon --name myCsAddon --template=cs
📁 Creating addon directory: myCsAddon
✅ C# addon 'myCsAddon' created successfully!
📁 D:\work\w\myCsAddon

Next steps:
  1. npm run build-myCsAddon
  2. See myCsAddon/README.md for usage examples
 ```

Here's how to use it from Javascript:
```js
// Load the C# module
const myCsAddon = require('./myCsAddon/build/Release/myCsAddon');

// Call exported methods (note: C# method names are converted to camelCase in JavaScript)
console.log(myCsAddon.Addon.hello('World'));
// Output: Hello from C#, World!

console.log(myCsAddon.Addon.add(5, 3));
// Output: 8

console.log(myCsAddon.Addon.getCurrentTime());
// Output: (current time)
```

Here's what the output looks like:
```
PS D:\work\w> node .\t.js
Hello from C#, World!
8
2025-10-22 17:36:15
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ npx winapp --help
 
 **Node.js/Electron Specific:**
 
-- [`node create-addon`](./docs/usage.md#node-create-addon) - Generate native C++ addons
+- [`node create-addon`](./docs/usage.md#node-create-addon) - Generate native C# or C++ addons
 - [`node add-electron-debug-identity`](./docs/usage.md#node-add-electron-debug-identity) - Add identity to Electron processes
 
 The full CLI usage can be found here: [Documentation](/docs/usage.md)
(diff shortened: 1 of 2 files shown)


src/winsdk-npm/cli.js
```diff
@@ -197,47 +201,77 @@ async function handleNode(args) {
...
     console.log('');
+    console.log('Templates:');
+    console.log('  cpp                   C++ native addon (node-gyp)');
+    console.log('  cs                    C# addon (node-api-dotnet)');
+    console.log('');
     console.log('Examples:');
     console.log(`  ${CLI_NAME} node create-addon`);
-    console.log(`  ${CLI_NAME} node create-addon --name myCustomAddon`);
+    console.log(`  ${CLI_NAME} node create-addon --name myAddon`);
+    console.log(`  ${CLI_NAME} node create-addon --template=cs --name MyCsAddon`);
```
JesseCol: I'm realizing we have the "--template" parameter using "=" and the "--name" parameter NOT using "=".  I'll change this to remove the "=" from the examples.

PR #152, opened 2025-11-19
When a tool fails, attempt to show lines that contain error information

JesseCol:
Before this change, when there's an error with a tool it's unclear why it's happening.  Users must know to re-run with "--verbose" option to see the error text.

After this change, here's what the output looks like when there's an error in MakeAppx:
```
PS D:\temp\rigvgl1m.zgh> winapp.exe pack .
[ERROR] - MakeAppx : error: Error info: error C00CE169: App manifest validation error: The app manifest must be valid as per schema: Line 38, Column 7, Reason: 'InvalidValue' violates enumeration constraint of 'appContainer mediumIL'.
[ERROR] - The attribute '{http://schemas.microsoft.com/appx/manifest/uap/windows10/10}TrustLevel' with value 'InvalidValue' failed to parse.
[ERROR] - MakeAppx : error: Package creation failed.
[ERROR] - MakeAppx : error: 0x80080204 - The specified package format is not valid: The package manifest is not valid.
[ERROR] - makeappx.exe reported an error. The above text is only part of the tool output. Please re-run with --verbose to see the full output.
❌ Failed to create MSIX package: Failed to create MSIX package: makeappx.exe execution failed with exit code 1
```

Here's what it looks like when there's an error in MakePri:
```
PS D:\temp\rigvgl1m.zgh> winapp.exe pack .
[ERROR] - ERROR: PRI191: 0x80004005 - Appx manifest not found or is invalid. Please ensure well-formed manifest file is present. Or specify an index name with /in switch.
End tag 'InvalidEndTag' does not match the start tag 'InvalidTag'.


[ERROR] - makepri.exe reported an error. The above text is only part of the tool output. Please re-run with --verbose to see the full output.
❌ Failed to create MSIX package: Failed to create MSIX package: Failed to generate PRI file: makepri.exe execution failed with exit code -2147467259
```

### DIFF
--- a/samples/electron/csAddon/csAddon.csproj
+++ b/samples/electron/csAddon/csAddon.csproj
@@ -1,6 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>net8.0-windows10.0.19041.0</TargetFramework>
+    <TargetFramework>net10.0-windows10.0.26100.0</TargetFramework>
     <OutputType>Library</OutputType>
     <AppendTargetFrameworkToOutputPath>false</AppendTargetFrameworkToOutputPath>
     <AssemblyName>csAddon</AssemblyName>
--- a/samples/electron/package-lock.json
+++ b/samples/electron/package-lock.json
@@ -449,7 +449,6 @@
       "integrity": "sha512-zx0EIq78WlY/lBb1uXlziZmDZI4ubcCXIMJ4uGjXzZW0nS19TjSPeXPAjzzTmKQlJUZm0SbmZhPKP7tuQ1SsEw==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "chalk": "^4.1.1",
         "fs-extra": "^9.0.1",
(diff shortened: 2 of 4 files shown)


src/winapp-CLI/WinApp.Cli/Services/BuildToolsService.cs
```diff
@@ -347,4 +397,11 @@ public InvalidBuildToolException(int processId, string stdout, string stderr, st
 
     [GeneratedRegex(@"^\d+\.\d+\.\d+\.\d+$")]
     private static partial Regex VersionFolderRegex();
+
+    // Recognize lines that look like error information.
+    // Example strings we want to match:
+    //   error:
+    //   general error c1010070:
+    [GeneratedRegex(@"error\s+[a-z]?\d+:|error:", RegexOptions.IgnoreCase)]
+    private static partial Regex ErrorLineRegex();
```
azchohfi: Should we filter the lines like this? I'm wondering if we just output everything if the return code is not zero. Wouldn't that be better? Maybe the tool being called outputs in multiple lines and the actual error would not be printed?

JesseCol: Hmm, I think I agree with you.  We might as well not add complexity if we don't need it.

I've noticed when coding with LLMs that they like short output, that's part of why I wanted to see if we could just show the error message here.  But I think you're right we don't really need that.

Issue #120, opened 2025-11-13
Need to Set a Custom location for .winapp


LanceMcCarthy:
Currently, the `.winapp` folder is placed in the user's directory.

However, on space constrained systems, these caches are usually placed on a different SSD. For example, I put all the caches I can on the D:/ or E:/ SSD drives as space on C: is a very high premium.

As a feature, I would like to have a way to set the cache directory, similar to how the nuget http and packages cache can be changed.

JesseCol:
You can actually already set the location of the global cache by setting the `WINAPP_CLI_CACHE_DIRECTORY` env var.

Granted, this doesn't appear to be documented anywhere.

PR #168, opened 2025-11-22
Add docs about Global Cache Directory

JesseCol:
Add docs about Global Cache Directory

Inspired by issue #120

### DIFF
--- a/.pipelines/release.yml
+++ b/.pipelines/release.yml
@@ -210,9 +210,10 @@ extends:
               condition: always()
               inputs:
                 ConnectedServiceName: ${{ parameters.signingIdentity.serviceName }}
-                usemanagedidentity: true
+                usemanagedidentity: false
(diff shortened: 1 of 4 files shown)


docs/usage.md
```diff
@@ -425,3 +427,25 @@ All commands support these global options:
...
+### Global Cache Directory
+
+Winapp creates a directory to cache files that can be shared between multiple projects.
+
+By default, winapp creates a directory at `$UserProfile/.winapp` as the global cache directory.
+
+To use a different location, set the `WINAPP_CLI_CACHE_DIRECTORY` environment variable:
+
+```cmd
+REM cmd - set a custom location for winapp's global cache
```
azchohfi: I would add these as headers to the "code" sections, instead of inside the comment/remark itself. It makes it clearer that the block is for cmd vs pwsh.

azchohfi (approved):

PR #173, opened 2025-11-27
Allow winapp init in subdirectory of user home dir

JesseCol:
If the user runs 'winapp init' in a dir where there's an ancestor dir that already has a subdir named `.winapp`, the operation fails.  This is by design.

The problem is when the ancestor dir contains a subdir named `.winapp` that is the global winapp cache directory.  In this case, we find that `.winapp` subdir and mistakenly assume that's the root dir of the winapp project.

- [x] Needs manual testing
- [ ] Needs automated test code


azchohfi (approved):

azchohfi:
I'm ok with this PR as-is.

PR #178, opened 2025-12-03
Change name of global cache dir to .winappglobal instead of .winapp

JesseCol:
We have a test case that's intermittently failing, it turns out this is because some of the tests set the global .winapp dir and local .winapp dir to be the same.  I added logic recently that assumed they could never be the same, which was a bad assumption.

To avoid confusion over these dirs, let's rename the global cache dir to .winappglobal.

### DIFF
--- a/samples/electron/package-lock.json
+++ b/samples/electron/package-lock.json
@@ -26,7 +26,7 @@
         "electron": "38.2.2",
         "nan": "^2.23.0",
         "node-addon-api": "^8.5.0",
-        "node-gyp": "^11.4.2"
+        "node-gyp": "^11.5.0"
       }
     },
     "../../src/winapp-npm": {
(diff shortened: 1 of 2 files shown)


src/winapp-CLI/WinApp.Cli/Services/WinappDirectoryService.cs
```diff
@@ -35,23 +35,31 @@ public DirectoryInfo GetGlobalWinappDirectory()
...
         {
             var winappDirectory = Path.Combine(dir.FullName, ".winapp");
-            if (Directory.Exists(winappDirectory) && !string.Equals(winappDirectory, globalWinappDirectory.FullName, StringComparison.OrdinalIgnoreCase))
+            if (Directory.Exists(winappDirectory))
             {
+                // Is this .winapp folder in the UserProfile directory?
+                var userProfile = Environment.GetFolderPath(Environment.SpecialFolder.UserProfile);
+                if (string.Equals(dir.FullName, userProfile, StringComparison.OrdinalIgnoreCase))
+                {
+                    throw new InvalidOperationException(
```
azchohfi: Should this be an error, or just a "warning"? They could be just trying to (weirdly, sure) run the command in their user profile folder. Maybe if we can find the `packages` folder inside the `.winapp` folder, then we know this is this scenario.

PR #185, opened 2025-12-09
Adding a CLI packaging guide

nmetulev:


docs/guides/packaging-cli.md
```diff
@@ -0,0 +1,166 @@
...
+```
+
+This creates a `devcert.pfx` file. To trust this certificate on your development machine, install it (requires administrator privileges):
+
+```powershell
+# Run PowerShell as Administrator
+winapp cert install
+```
+
+**Important**: Keep your development certificate outside the folder containing your CLI executable to avoid accidentally including it in the package.
```
JesseCol, line 149: I didn't know it was that easy to accidentally include devcert.pfx in the package.  Should we add logic to winappCLI to detect this and block?

JesseCol (approved):

docs/guides/packaging-cli.md
nmetulev: That's a good idea. Should we block though or just warn? Let me create an issue where we can discuss

PR #186, opened 2025-12-09
Improve the user experience for migrating from .winapp to .winappglobal

JesseCol:
We're changing the default global dir from $UserProfile/.winapp to $UserProfile/.winappglobal.  

We want to make it easy and not consume too much hard drive space.  The idea in this PR is to still honor the old location for now, and then we'll be more aggressive about using the new location in the future.

See [specs/move-global-folder-spec.md](https://github.com/microsoft/WinAppCli/compare/jessecol/plan-for-migratating-to-winappglobal?expand=1#diff-74933c3ee285390746d3cbf27f7d7ead71dcbe4b3f50f070ed23e79a1995a4b9) for more information.

Here's how a restore looks:
image


JesseCol:
Another option -- we could just go back to using ".winapp" as the global folder and find some other way to distinguish it from a local ".winapp" folder.  @nmetulev @azchohfi

PR #187, opened 2025-12-09
Added a banner when a user calls just winapp

nmetulev:
image


nmetulev:
Thinking about your earlier comment too about how the command is just winapp, how does this look instead 
image

JesseCol:
> Thinking about your earlier comment too about how the command is just winapp, how does this look instead image

Love it!

JesseCol:
nit: description has image of old banner

PR #195, opened 2025-12-16
Improved logs, and improved init command.

### LINKED ISSUE #64: Improve Init
- [ ] init should ask for project type
- [ ] allow override on PrompForValue (show default, enter accepts, anything else replaces)
- [ ] generate manifest shouldn't ask questions
- [ ] electron should override initial question for project type (first question, node, python, exe(default))
- [ ] init should ask if you want to install the sdks (both in a single question)

### LINKED ISSUE #179: Rethink verbosity and developer experience for all commands
Today, most commands spit out a various log of work performed, spamming the output, even when verbosity is set to default. This makes it dificult to understand what each command does spams the console window.

To fix this:
1. all commands should print the bare minimum by default. A single line for most will be enough
2. We should use progress rings that show the details of work happening, and then clearing the details to indicate the work was completed
3. show relevant errors and logs only if something fails. if something fails, we should then hint to the user to run the command with more verbosity to get more logs or details

For example, the init command could have the following stages:

## Init example

Starting initialization
```bash
> winapp init
  ✔ Created config file: D:\temp\cli-temp\app1\winapp.yaml
  ❯ Installing SDK dependencies (WinAppSDK, WindowsSDK, CppWinRT, BuildTools)
      ✔ Microsoft.WindowsAppSDK 1.8.251106002 downloaded
      ⠸ Downloading Microsoft.Windows.ImplementationLibrary
```

Once packages are downloaded
```bash
> winapp init
  ✔ Created config file: D:\temp\cli-temp\app1\winapp.yaml
  ✔ Installed SDK dependencies (WindowsAppSDK, WindowsSDK, CppWinRT, BuildTools)
  ❯ Generating C++/WinRT projections
      ✔ Generated C++/WinRT projections with cppwinrt
      ⠸ Copying headers to D:\temp\cli-temp\app1\.winapp\inlcude
```

Completed initialization
```bash
> winapp init
  ✔ Created config file: .\winapp.yaml
  ✔ Installed SDK dependencies (WindowsAppSDK, WindowsSDK, CppWinRT, BuildTools)
  ✔ Generated C++/WinRT projections in .\.winapp
  ✔ Created manifest: .\appxmanifest.xml
  ✔ Created dev certificate for package signing: .\devcert.pfx
  ✔ Installed Windows App Runtime
  ✔ Enabled dev mode
  ✔ Installed Visual Studio build tools
  ✔ Added .winapp and devcert.pfx to .gitignore
>
```

## Commands that need to be updated:

- [ ] init
- [ ] restore
- [ ] pack
- [ ] update
- [ ] `create-debug-identity`
- [ ] `node add-electron-debug-identity`
- [ ] `node create-addon` all templates

azchohfi:
Fixes #179
Fixes #64

### DIFF
--- a/src/winapp-CLI/WinApp.Cli/Commands/ManifestUpdateAssetsCommand.cs
+++ b/src/winapp-CLI/WinApp.Cli/Commands/ManifestUpdateAssetsCommand.cs
@@ -65,7 +65,7 @@ internal class ManifestUpdateAssetsCommand : Command
                 try
                 {
                     await manifestService.UpdateManifestAssetsAsync(manifestPath, imagePath, taskContext, cancellationToken);
-                    return (0, $"{UiSymbols.Check} Successfully updated assets in manifest.");
+                    return (0, "Successfully updated assets for manifest.");
                 }
                 catch (Exception ex)
                 {
--- a/src/winapp-CLI/WinApp.Cli/Commands/SignCommand.cs
+++ b/src/winapp-CLI/WinApp.Cli/Commands/SignCommand.cs
@@ -61,7 +61,7 @@ internal class SignCommand : Command
                 {
                     await certificateService.SignFileAsync(filePath, certPath, taskContext, password, timestamp, cancellationToken);
 
-                    return (0, $"{UiSymbols.Lock} Signed file: {filePath}");
+                    return (0, "Signed file: {filePath}");
                 }
                 catch (InvalidOperationException error)
                 {
--- a/src/winapp-CLI/WinApp.Cli/Commands/UpdateCommand.cs
+++ b/src/winapp-CLI/WinApp.Cli/Commands/UpdateCommand.cs
@@ -149,7 +149,7 @@ internal class UpdateCommand : Command
                         taskContext.AddDebugMessage($"{UiSymbols.Note} Windows App SDK packages not found, skipping runtime installation");
                     }
 
-                    return (0, $"{UiSymbols.Party} Update completed successfully!");
+                    return (0, "Update completed successfully!");
                 }
                 catch (Exception error)
                 {
--- a/src/winapp-CLI/WinApp.Cli/ConsoleTasks/GroupableTask.cs
+++ b/src/winapp-CLI/WinApp.Cli/ConsoleTasks/GroupableTask.cs
@@ -111,9 +111,17 @@ internal class GroupableTask<T> : GroupableTask
     {
         if (task.IsCompleted)
         {
-            static string FormatCheckMarkMessage(string indentStr, string? message)
+            static string FormatCheckMarkMessage(string indentStr, string message)
             {
-                return $"{indentStr}[green]{Emoji.Known.CheckMarkButton}[/] {message}";
+                bool firstCharIsEmojiOrOpenBracket = false;
+                if (message.Length > 0)
+                {
+                    var firstChar = message[0];
+                    firstCharIsEmojiOrOpenBracket = char.IsSurrogate(firstChar)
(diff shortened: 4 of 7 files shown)


nmetulev (changes requested):
Amazing, here are my suggestions for improvement:

All questions that can be asked prior to doing work should be asked. Right now the first question is do you want to install the sdks and it immediately goes through installing sdks, and then it starts asking questions on package name, etc. It should ask all those questions first - installing sdks should not be the first question.

Also, as the questions are asked, the answer to those questions should stay visible. for example:

```bash

Package name: my app
Publisher: bla
Version: bla
Setup SDKs: (Windows SDK 10.0.28000.1-rtm, WinAppSDK 2.0.250930001-experimental1)
............

✅ Setup completed successfully
  ✅ Workspace initialized
  ✅ Developer mode: enabled
  ✅ SDK packages downloaded and C++ headers generated in .winapp
  ✅ WinAppSDK Runtime installed: 2.0.250930001-experimental1
  ✅ Manifest and Assets created: appxmanifest.xml
  ✅ Configuration file created: winapp.yaml
  ✅ Development certificate created: devcert.pfx
  ✅ Added .winapp and devcert.pfx to .gitignore

```

Other things:
- [ ] Change `Install SDKs` to `Setup SDKs`
- [ ] Change `SDK packages downloaded and C++ headers generated in .winapp` to include WASDK (`SDK and WASDK packages downloaded and C++ headers generated in .winapp`)
- [ ] Calling init that already has a config file present still asks for what SDK version to install - should just install what's in the config unless --ignore-config is passed

nmetulev:
Thank you. I think we are very close. Few minor things:

image

* (red line) Remove the default choice once the user hits enter on the prompts so it's not duplicated
* (yellow arrow) underline the answer for each prompt just like you have it for the setup piece so it matches
* (blue line) Add a new line to separate the questions from setup

image

* Remove the new line between the question and the options

nmetulev:
And there is a bug, when I go through init and when i select not to setup the sdks, the entire init process is canceled. I'd expect the manifest, cert, etc to still be created 
image

azchohfi:
The only thing I haven't done yet is:
"Make it clear that the setup here is for C++, change to "Setup SDKs for C++ development:""

I agree that it is mostly for that, but it is not only for it. The steps that generate C++/WinRT projections are much later in the process too. This is so we download the WinSdk and WinAppSdk NuGets, so we have a lot of things available, only one of which is the WinMDs for C++/WinRT projections. We need them for WinAppSDK Runtime, for self-contained deployment too. This also allows any language to have access to all their dlls. I'm not against changing this text, but we need to be clear. I just want to avoid "scaring" developers who don't think they use C++, even if we say "C++", since they might need it down the line, like Electron developers. Does that make sense?

nmetulev:
> The only thing I haven't done yet is: "Make it clear that the setup here is for C++, change to "Setup SDKs for C++ development:""
> 
> I agree that it is mostly for that, but it is not only for it. The steps that generate C++/WinRT projections are much later in the process too. This is so we download the WinSdk and WinAppSdk NuGets, so we have a lot of things available, only one of which is the WinMDs for C++/WinRT projections. We need them for WinAppSDK Runtime, for self-contained deployment too. This also allows any language to have access to all their dlls. I'm not against changing this text, but we need to be clear. I just want to avoid "scaring" developers who don't think they use C++, even if we say "C++", since they might need it down the line, like Electron developers. Does that make sense?

Good point. it sounds like we need to separate the "download and setup all dependencies" and "generate C++/WinRT headers" steps, to that the first one happens regardless, and the second is optional?

nmetulev (approved):

PR #214, opened 2026-01-06
Enable running Microsoft Store Developer CLI commands via a new 'store' subcommand.

### LINKED ISSUE #142: Proxy msstore CLI commands
It would be pretty cool to have the `winapp` CLI be able to proxy commands over to the `msstore` CLI. 

How would this look? How about using the same approach that dotnet CLI handles workloads, with a command subset.

- dotnet **workload <workload command>**

- winapp store <[msstore CLI command](https://learn.microsoft.com/en-us/windows/apps/publish/msstore-dev-cli/overview#commands)>

For example:
```pwsh
winapp store --help
winapp store init
winapp store package
winapp store publish
```
😍

Now, about implementation, I don't imagine that you're want to completely rebuild the work that Alex is doing over in the msstore CLI repo, two thoughts:

- Just make the msstore CLI package a dependency and proxy the commands
- submodule the [msstore cli repo](https://github.com/microsoft/msstore-cli) and do the compile/build yourself and implement the commands

azchohfi:
Fixes #142

### DIFF
--- a/.github/actions/collect-metrics/action.yml
+++ b/.github/actions/collect-metrics/action.yml
@@ -1,5 +1,5 @@
 name: Collect Build Metrics
-description: Measure artifact sizes, build warnings, test results, and CLI startup time
+description: Measure artifact sizes, test results, and CLI startup time
 
 inputs:
   solution-path:
@@ -35,37 +35,6 @@ runs:
         Write-Host "Current artifact sizes:"
         $sizes.GetEnumerator() | Sort-Object Key | ForEach-Object {
           Write-Host "  $($_.Key): $([math]::Round($_.Value / 1MB, 2)) MB ($($_.Value) bytes)"
-        }
-
-    - name: Count build warnings
-      shell: pwsh
-      run: |
-        # Re-run build (incremental, near-instant) to capture warning output
-        # We only care about the warnings in stdout, not the exit code
-        $output = dotnet build "${{ inputs.solution-path }}" -c Release 2>&1 | Out-String
-        # Ignore build exit code — we're just scraping warnings from the output
-        $global:LASTEXITCODE = 0
-        $warningLines = ($output -split "`n") | Where-Object { $_ -match ': warning [A-Z]{2,}\d+:' }
-        $warningCount = ($warningLines | Measure-Object).Count
-
-        # Categorize warnings
-        $trimWarnings = ($warningLines | Where-Object { $_ -match ': warning IL\d+:' } | Measure-Object).Count
-        $codeAnalysis = ($warningLines | Where-Object { $_ -match ': warning CA\d+:' } | Measure-Object).Count
-        $otherWarnings = $warningCount - $trimWarnings - $codeAnalysis
-
-        $warnings = @{
-          total = $warningCount
-          trimming = $trimWarnings
-          codeAnalysis = $codeAnalysis
-          other = $otherWarnings
-        }
-
-        $warnings | ConvertTo-Json | Set-Content "${{ inputs.artifacts-path }}/build-warnings.json"
-
-        Write-Host "Build warnings: $warningCount total ($trimWarnings trimming/AOT, $codeAnalysis code analysis, $otherWarnings other)"
-        if ($warningLines) {
-          Write-Host "`nWarning details:"
-          $warningLines | ForEach-Object { Write-Host "  $_" }
         }
 
     - name: Parse test results
--- a/.github/actions/report-metrics/action.yml
+++ b/.github/actions/report-metrics/action.yml
@@ -29,7 +29,6 @@ runs:
       run: |
         New-Item -ItemType Directory -Path "metrics-baseline" -Force | Out-Null
         Copy-Item "${{ inputs.artifacts-path }}/binary-sizes.json" "metrics-baseline/sizes.json" -ErrorAction SilentlyContinue
-        Copy-Item "${{ inputs.artifacts-path }}/build-warnings.json" "metrics-baseline/warnings.json" -ErrorAction SilentlyContinue
         Copy-Item "${{ inputs.artifacts-path }}/test-summary.json" "metrics-baseline/tests.json" -ErrorAction SilentlyContinue
         Copy-Item "${{ inputs.artifacts-path }}/startup-time.json" "metrics-baseline/startup.json" -ErrorAction SilentlyContinue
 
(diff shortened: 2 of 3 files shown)


src/winapp-CLI/WinApp.Cli/Services/MSStoreCLIService.cs
```diff
@@ -0,0 +1,71 @@
...
+        if (!IsMSStoreCLIInPath())
+        {
+            // MSStoreCLI is not in the path, proceed to install it via winget
+            var confirm = Program.PromptYesNo("MSStoreCLI not installed - install MSStore Developer CLI with winget (user interaction may be required)? (y/N) ");
+            if (!confirm)
+            {
+                throw new InvalidOperationException("MSStoreCLI is required but not installed.");
+            }
+
+            var args = "install \"Microsoft Store Developer CLI\" --source msstore";
```
nmetulev: Should we use the id here instead?

src/winapp-CLI/WinApp.Cli/Services/MSStoreCLIService.cs
```diff
@@ -0,0 +1,71 @@
...
+            logger.LogDebug("Running winget with arguments: {Args}", args);
+
+            ProcessStartInfo installProcessStartInfo = new()
+            {
+                FileName = "winget",
+                Arguments = "install \"Microsoft Store Developer CLI\" --source msstore",
```
nmetulev: args should be reused from above

nmetulev (changes requested):
Awesome. before merging:
- [ ] Need to have msstore team review and approve
- [ ] Usage docs updated to include store command and link to msstore docs
- [ ] Readme updated to include store command in the list of commands
- [ ] Add tests
- [ ] Should we have a silent install option (install MSStoreCLI if not installed) for things like CI? Or a command to ensure it is installed prior to calling it `winapp store ensure-available`? The second option seems better to me.

src/winapp-CLI/WinApp.Cli/Commands/MSStoreCommand.cs
```diff
@@ -0,0 +1,59 @@
...
+using Microsoft.Extensions.Logging;
+using System.CommandLine;
+using System.CommandLine.Invocation;
+using WinApp.Cli.Services;
+
+namespace WinApp.Cli.Commands;
+
+internal class MSStoreCommand : Command
+{
+    public MSStoreCommand() : base("store", "Run a Microsoft Store Developer CLI command.")
```
nmetulev: We should make it clear that it requires an additional cli to be installed. Maybe we add an aka.ms link for more info here?

src/winapp-CLI/WinApp.Cli/Services/MSStoreCLIService.cs
azchohfi: We can, its just a weird command since its not a readable ID (its the store big id):
```
winget install --id 9P53PC5S0PHJ -e --source msstore
```

src/winapp-CLI/WinApp.Cli/Commands/MSStoreCommand.cs
azchohfi: Done.

azchohfi:
> * [ ]  Should we have a silent install option (install MSStoreCLI if not installed) for things like CI? Or a command to ensure it is installed prior to calling it `winapp store ensure-available`? The second option seems better to me.

We have 3 options:
1) Assume that the command we have is interactive, since it requires accepting the store's EULA (through winget). We are using the winget's "Microsoft Store" source, and that source requires accepting the EULA at first usage.
Not sure how that works in CI though. Also, this is a solved problem (msstore cli installation in CI), that we might not want to solve again (https://github.com/microsoft/microsoft-store-apppublisher). Its a single line in their yaml file.:
```
    - uses: microsoft/microsoft-store-apppublisher@v1.2
```
But we are passing that burden to them, which is not ideal.

2) Add a `--silent`, or something around these lines, that will just force installation by passing --acept-soure-agreements to winget. This might work.

3) We change the installation method to be similar to what https://github.com/microsoft/microsoft-store-apppublisher does, which means moving away from using winget and using the GitHub release to download the zip of a specific version. We then extract that folder to our "tools" folder within the global `.winapp` folder, which is a known folder for us, so we just call it. Would also work fine, and this is probably the more complete solution, but is way more work on us.

docs/cli-schema.json
```diff
@@ -1098,6 +1098,44 @@
         }
       }
     },
+    "store": {
+      "description": "Run a Microsoft Store Developer CLI command. This command might prompt the installation of Microsoft Store Developer CLI (https://aka.ms/msstoredevcli).",
```
nmetulev, line 1089: We shouldn't prompt, we should pull it down automatically like we do built tools

docs/cli-schema.json
```diff
@@ -1098,6 +1098,44 @@
...
+      "hidden": false,
+      "options": {
+        "--quiet": {
```
nmetulev: we shouldn't mess with the output of the ms store cli, right, do we need this one?

docs/cli-schema.json
```diff
@@ -1098,6 +1098,44 @@
...
+          "hasDefaultValue": true,
+          "defaultValue": false,
+          "arity": {
+            "minimum": 0,
+            "maximum": 1
+          },
+          "required": false,
+          "recursive": false
+        },
+        "--verbose": {
```
nmetulev: same here, wouldn't the ms store cli have their own --verbose flag

src/winapp-CLI/WinApp.Cli/Services/MSStoreCLIService.cs
```diff
@@ -0,0 +1,96 @@
...
+internal class MSStoreCLIService(IAnsiConsole ansiConsole, IWinappDirectoryService winappDirectoryService, ILogger<MSStoreCLIService> logger) : IMSStoreCLIService
+{
+    private static readonly HttpClient Http = new();
+    private const string ExeName = "msstore.exe";
+
+    public async Task EnsureMSStoreCLIAvailableAsync(CancellationToken cancellationToken = default)
+    {
+        if (!IsMSStoreCLIAvailable())
+        {
+            var confirm = await ansiConsole.PromptAsync(new ConfirmationPrompt("MSStoreCLI not installed - download and install MSStore Developer CLI?"), cancellationToken);
```
nmetulev: Since we are managing the binary, we should just do it like we do the build tools and not prompt. We are not installing it so we should change the language to not say install throughout

nmetulev (changes requested):

docs/cli-schema.json
azchohfi: We are forwarding this, but we also need ours, in case the user wants verbose, for example, on the download info.

azchohfi: I will remove it from our docs since it doesn't make much sense anyway.

PR #246, opened 2026-01-21
Adding electron e2e test

### LINKED ISSUE #241: Create an end to end test for Electron
We need an end to end test that executes the following flow:

* create new electron app
* run init
* create cs addon
* build cs addon
* create c++ addon
* build c++ addon
* run add-electron-debug-identity
* npm start to ensure app starts
* package app with npx winapp app

This will help us avoid regressions as we continue adding features.

nmetulev:
Fixes #241

### DIFF
--- a/src/winapp-npm/src/msix-utils.ts
+++ b/src/winapp-npm/src/msix-utils.ts
@@ -21,10 +21,16 @@ export interface ElectronDebugIdentityResult {
   assetsDir: string;
 }
 
+export interface ClearElectronDebugIdentityResult {
+  success: boolean;
+  electronExePath: string;
+  restoredFromBackup: boolean;
+}
+
 /**
- * Adds MSIX identity information from an appxmanifest.xml file to an executable's embedded manifest
+ * Adds package identity information from an appxmanifest.xml file to an executable's embedded manifest
  * @param exePath - Path to the executable file
- * @param appxManifestPath - Path to the appxmanifest.xml file containing MSIX identity data
+ * @param appxManifestPath - Path to the appxmanifest.xml file containing package identity data
  * @param options - Optional configuration
  */
 export async function addMsixIdentityToExe(
@@ -35,7 +41,7 @@ export async function addMsixIdentityToExe(
   const { verbose = false } = options;
 
   if (verbose) {
-    console.log('Adding MSIX identity to executable using native CLI...');
+    console.log('Adding package identity to executable using native CLI...');
   }
 
   // Build arguments for native CLI
@@ -62,7 +68,7 @@ export async function addMsixIdentityToExe(
 }
 
 /**
- * Adds MSIX identity to the Electron debug process
+ * Adds package identity to the Electron debug process
  * @param options - Configuration options
  */
 export async function addElectronDebugIdentity(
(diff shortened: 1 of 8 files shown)


src/winapp-npm/src/cs-addon-utils.ts
```diff
@@ -289,12 +289,20 @@ async function installNodeApiDotnet(projectRoot: string, verbose: boolean): Prom
     console.log(`📦 Installing node-api-dotnet...`);
   }
 
-  const installCommand = 'npm install node-api-dotnet';
+  // Get npm path - use the npm that's in the same location as the running node
+  const nodeDir = path.dirname(process.execPath);
+  const npmCmd = process.platform === 'win32' ? 'npm.cmd' : 'npm';
```
azchohfi, line 294: Same here.

src/winapp-npm/src/cpp-addon-utils.ts
```diff
@@ -166,10 +166,21 @@ async function installRequiredPackages(projectRoot: string, verbose: boolean): P
       console.log(`📦 Installing missing packages: ${missingPackages.join(', ')}`);
     }
 
-    const installCommand = `npm install --save-dev ${missingPackages.join(' ')}`;
+    // Get npm path - use the npm that's in the same location as the running node
+    const nodeDir = path.dirname(process.execPath);
+    const npmCmd = process.platform === 'win32' ? 'npm.cmd' : 'npm';
```
azchohfi, line 171: This code path is dead, why are we checking for win32 vs not?

src/winapp-npm/src/msix-utils.ts
```diff
@@ -108,6 +110,12 @@ export async function addElectronDebugIdentity(
 
     // Build arguments for native CLI
     const args = ['create-debug-identity', electronExePath];
+    if (manifest) {
+      args.push('--manifest', manifest);
+    }
+    if (noInstall) {
+      args.push('--no-install');
```
azchohfi, line 172: We've added these just for completeness, right? no-install was false already and we are just sending it as false.

scripts/test-e2e-electron.ps1
```diff
@@ -0,0 +1,466 @@
...
+    # Create Electron Application
+    # ========================================================================
+
+    Write-TestHeader "E2E Electron Test - Create Electron App Phase"
+
+    Write-TestStep "Creating new Electron app..." 3
+
+    # Clear npm cache to avoid ECOMPROMISED errors in CI environments
+    Write-Verbose "Clearing npm cache to avoid lock file issues..."
+    npm cache clean --force 2>$null
```
azchohfi: This is quite destructive... Do we need it? It will make restoring packages on separate projects very slow since it will need to download it again. Not that we will be running these tests locally that much since it's not called from the main `build-cli.ps1` but would be better to avoid this. Also not sure if this matters since we are setting `npm_config_cache`. Oh, actually, we have this exact line inside the while loop, so we don't need this here. Still, same comment applies to the line inside the while loop.

scripts/test-e2e-electron.ps1
```diff
@@ -0,0 +1,466 @@
...
+    while (-not $electronAppCreated -and $retryCount -lt $maxRetries) {
+        $retryCount++
+        Write-Verbose "Attempt $retryCount of $maxRetries to create Electron app..."
+        
+        try {
+            # Use --prefer-offline to reduce network issues and clear package-lock before retry
+            if ($retryCount -gt 1) {
+                Write-Verbose "Cleaning up failed attempt..."
+                Remove-Item -Path (Join-Path $testDir "electron-app") -Recurse -Force -ErrorAction SilentlyContinue
+                npm cache clean --force 2>$null
```
azchohfi, line 238: Here.

scripts/test-e2e-electron.ps1
```diff
@@ -0,0 +1,466 @@
...
+    $addIdentityCommand = "npx winapp node add-electron-debug-identity --no-install"
+    Assert-Command $addIdentityCommand "Failed to add Electron debug identity"
+
+    # Verify manifest was updated
+    $manifestPath = Join-Path $electronAppDir "appxmanifest.xml"
+    $manifestContent = Get-Content $manifestPath -Raw
+    if ($manifestContent -like "*VisualElements*") {
+        Write-TestSuccess "Manifest updated with visual elements"
+    }
+
```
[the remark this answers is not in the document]
nmetulev: I tried to get this to work but it is not reliable enough in a CI environment and it doesn't really tell us anything. let's skip it

scripts/test-e2e-electron.ps1
```diff
@@ -0,0 +1,466 @@
...
+    $installCommand = "npm install $localNpmPackagePath --save-dev"
+    Assert-Command $installCommand "Failed to install winapp npm package"
+
+    # Verify winapp is installed
+    $nodeModulesPath = Join-Path $electronAppDir "node_modules" ".bin" "winapp"
+    if ($PSVersionTable.Platform -eq 'Win32NT' -or $PSVersionTable.OS -like 'Windows*') {
+        $winappCli = Join-Path $electronAppDir "node_modules" ".bin" "winapp.cmd"
+        Assert-FileExists $winappCli "winapp CLI"
+    } else {
+        Assert-FileExists $nodeModulesPath "winapp CLI"
```
azchohfi: We don't need this. It's always Windows.

scripts/test-e2e-electron.ps1
```diff
@@ -0,0 +1,466 @@
...
+    Write-TestStep "Adding Electron debug identity..." 11
+
+    $addIdentityCommand = "npx winapp node add-electron-debug-identity --no-install"
+    Assert-Command $addIdentityCommand "Failed to add Electron debug identity"
+
+    # Verify manifest was updated
+    $manifestPath = Join-Path $electronAppDir "appxmanifest.xml"
+    $manifestContent = Get-Content $manifestPath -Raw
+    if ($manifestContent -like "*VisualElements*") {
+        Write-TestSuccess "Manifest updated with visual elements"
```
azchohfi: What does `add-electron-debug-identity` has to do with visual elements? Not sure I follow.

scripts/test-e2e-electron.ps1
```diff
@@ -0,0 +1,466 @@
...
+            if ($retryCount -gt 1) {
+                Write-Verbose "Cleaning up failed attempt..."
+                Remove-Item -Path (Join-Path $testDir "electron-app") -Recurse -Force -ErrorAction SilentlyContinue
+                npm cache clean --force 2>$null
+                Start-Sleep -Seconds 2
+            }
+            
+            $electronCommand = "npx -y create-electron-app@latest electron-app --template=webpack"
+            Write-Verbose "Running: $electronCommand"
+            Invoke-Expression $electronCommand
```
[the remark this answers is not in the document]
nmetulev: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/microsoft/winappCli/pull/246#discussion_r2714576658)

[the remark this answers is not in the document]
azchohfi: Not sure I want to pin... I understand the security risk, but it would be nice to check with latest to see if they break us somehow.

azchohfi (changes requested):

src/winapp-npm/src/msix-utils.ts
nmetulev: We were not sending --no-install, I needed to add it because CI doesn't have dev mode so when running the command, I wanted to skip install

scripts/test-e2e-electron.ps1
nmetulev: I needed to add it because it would sometimes fail to restore/install. But I can remove it from here

scripts/test-e2e-electron.ps1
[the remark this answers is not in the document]
nmetulev: Same, that's why I had it like that at first - though the security concerns are real and we should try to avoid any potential supply chain attacks.

scripts/test-e2e-electron.ps1
nmetulev: done

scripts/test-e2e-electron.ps1
nmetulev: Nothing

src/winapp-npm/src/msix-utils.ts
azchohfi: I see. Makes sense.

azchohfi (approved):

PR #248, opened 2026-01-21
fixed a bug where electron.backup.exe was getting poisoned and added a clear-electron-debug-identity command

nmetulev:


src/winapp-npm/src/msix-utils.ts
```diff
@@ -73,31 +79,80 @@ export async function addElectronDebugIdentity(
...
+        // Ignore errors reading version file
+      }
+    }
+
+    const backupExists = fsSync.existsSync(electronBackupPath);
+    let versionMismatch = false;
+    if (backupExists) {
+      if (currentElectronVersion && backupVersion) {
+        versionMismatch = currentElectronVersion !== backupVersion;
+      } else if (currentElectronVersion && !backupVersion) {
```
azchohfi:
```suggestion
    if (backupExists && currentElectronVersion) {
      if (backupVersion) {
        versionMismatch = currentElectronVersion !== backupVersion;
      } else {
```

src/winapp-npm/src/msix-utils.ts
```diff
@@ -140,3 +195,67 @@ export async function addElectronDebugIdentity(
     throw new Error(`Failed to add Electron debug identity: ${err.message}`);
   }
 }
+
+/**
+ * Clears/removes package identity from the Electron debug process by restoring from backup
+ * @param options - Configuration options
+ */
+export async function clearElectronDebugIdentity(
```
azchohfi, line 203: One improvement would be for this to uninstall the package.

nmetulev: Right, i think that's a bigger feature though as we should probably add it to the main cli first and then to this, let's leave it out for now

Issue #271, opened 2026-02-03
[Bug]: Share NuGet cache


azchohfi:
### Describe the bug

Today, the CLI has its own nuget cache for all nuget packages that we need (WinAppSDK, build tools, etc). We should not have our own cache, to save disk space.

### To Reproduce

1. Use the WinApp CLI on a dotnet project
2. See packages in C:\Users\{USER}\.nuget\packages
3. See C:\Users\{USER}\.winapp\packages

### Expected behavior

We should not have a global cache that is different from the per user nuget cache. We should also use the same environment variable set by the user for nuget(NUGET_PACKAGES)

PR #285, opened 2026-02-11
Add .NET project support to WinApp CLI

### LINKED ISSUE #284: [Feature]: Native .NET (.csproj) support in winapp cli for .net applications

Currently, winapp cli manages its own NuGet packages, which works well for frameworks that do not depend on NuGet (like Electron or Rust). However, for .NET projects such as WinUI and WPF, these projects manage their own dependencies via csproj/nuget. It's problematic for winapp cli to try to handle or generate extra projections or manage NuGets for these projects, as it can lead to duplication and confusion. Supporting .NET applications (WinUI/WPF/etc.) natively would provide a much better and more integrated experience, avoiding conflicts and redundant configuration.

To support .NET projects natively:
- When initializing, detect the presence of a .csproj file. If present, do NOT generate a winapp.yaml file.
- If the user chooses to install the WinApp SDK:
  - Check if the project's TargetFramework is supported. If not, offer to change it; if refused, show an error.
  - If accepted, update the TargetFramework in the csproj.
  - Add necessary NuGet references (WinAppSDK & BuildTools) directly to the csproj.
- For the `package` and `create-debug-identity` commands, use information from the csproj file (in the same way winapp.yaml is used today) to infer and manage the WinAppSDK runtime for manifests, packaging, and related tasks.

Overall, winapp cli should treat csproj as a first-class configuration source for .NET projects, completely avoiding winapp.yaml in this context to prevent duplication.

### Additional context

- This enables better alignment with Visual Studio and .NET project workflows.
- Reduces manual and error-prone configuration steps for .NET developers.
- Avoids excessive/duplicate configuration files and NuGet projections.
- Streamlines winapp cli adoption for the .NET ecosystem.

This spec targets WinUI, WPF, and similar .NET project types. Other framework types (Rust, Electron, Python, etc.) should continue to use winapp.yaml as usual.

azchohfi:
## Description

Adds .NET project support for WinAppCLI.

## Related Issue

Fixes #284, and #271

## Type of Change

- ✨ New feature

## Checklist

- [X] [docs/usage.md](../docs/usage.md) updated (if CLI commands changed)
- [X] [Language-specific guides](../docs/guides) updated (if applicable)
- [X] [Sample projects updated](../samples) to reflect changes (if applicable)

### DIFF
--- a/src/winapp-CLI/WinApp.Cli.Tests/PackageCommandTests.cs
+++ b/src/winapp-CLI/WinApp.Cli.Tests/PackageCommandTests.cs
@@ -333,6 +333,9 @@ public class PackageCommandTests : BaseCommandTests
         Assert.IsTrue(File.Exists(externalManifestPath), "External manifest should still exist");
         Assert.IsTrue(File.Exists(Path.Combine(externalAssetsDir, "Logo.png")), "External Logo.png should still exist");
         Assert.IsTrue(File.Exists(Path.Combine(externalAssetsDir, "StoreLogo.png")), "External StoreLogo.png should still exist");
+
+        // Verify the input folder was not polluted with a manifest copy
+        Assert.IsFalse(File.Exists(Path.Combine(packageDir, "AppxManifest.xml")), "Input folder should not contain AppxManifest.xml after packaging");
     }
 
     [TestMethod]
--- a/src/winapp-CLI/WinApp.Cli/Services/MsixService.cs
+++ b/src/winapp-CLI/WinApp.Cli/Services/MsixService.cs
@@ -711,8 +711,6 @@ internal partial class MsixService(
 
         // Update manifest content to ensure it's either referencing Windows App SDK or is self-contained
         manifestContent = await UpdateAppxManifestContentAsync(manifestContent, null, null, sparse: false, selfContained: selfContained, taskContext, cancellationToken);
-        var updatedManifestPath = Path.Combine(inputFolder.FullName, "appxmanifest.xml");
-        await File.WriteAllTextAsync(updatedManifestPath, manifestContent, Encoding.UTF8, cancellationToken);
 
         if (string.IsNullOrWhiteSpace(finalPackageName) || string.IsNullOrWhiteSpace(extractedPublisher))
         {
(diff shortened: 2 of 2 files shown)


src/winapp-CLI/WinApp.Cli/ConsoleTasks/GroupableTask.cs
```diff
@@ -56,14 +56,26 @@ public GroupableTask(string inProgressMessage, GroupableTask? parent, Func<TaskC
...
+            T? result = default;
+            if (result is ValueTuple<int, string> v)
+            {
+                v.Item1 = 1;
+                CompletedMessage = (T?)(object)v;
+            }
+            else
+            {
+                return result;
+            }
```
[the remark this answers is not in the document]
azchohfi: No, this is wrong. ValueTuple is a reference type, so it does work. It gives an empty instance. This is just for completeness, so it doesn't matter much. This should never happen.

nmetulev (changes requested):
The pr changes the order of the init command and asks for sdks installation first. Please revert it to the order it was before (name, description, etc first)

docs/guides/dotnet.md
```diff
@@ -66,6 +58,11 @@ catch (InvalidOperationException)
 }
 ```
 
+> **Note:** The `TargetFramework` will be updated automatically by `winapp init` in the next step. If you prefer to set it manually, open `dotnet-app.csproj` and change the `TargetFramework` to include the Windows SDK version:
```
nmetulev, line 61: This is a bit weird - they still have to do this for this step to work, so it's not really optional at this point. Should we instead move running the init command first, or we keep the original flow where they change it themselves even if the init command makes that change?

azchohfi: Lets move it up.

azchohfi: Wait, they still need to run it, sorry. Yeah, they should run without identity before running init.

azchohfi: Oh, sorry, it doesn't work this way "yet".

docs/guides/dotnet.md
```diff
@@ -87,17 +84,22 @@ winapp init
 ```
 
 When prompted:
+- **TargetFramework update**: If your `TargetFramework` doesn't include a supported Windows SDK version, you'll be prompted to update it (e.g., to `net10.0-windows10.0.26100.0`)
```
nmetulev, line 87: this is out of order now

PR #291, opened 2026-02-13
`winapp pack` Should Copy Input Folder, Appxmanifest, and Assets to Temp Directory

### LINKED ISSUE #281: [Bug]: Target package folder gets polluted and outdated assets reused when using package command without --manifest
### Describe the bug

When running the `package` command in winappCli without specifying the `--manifest` parameter, the tool searches for `appxmanifest.xml` in order of: the target package location, then the current directory. If a manifest or other asset already exists in the package target directory (from a previous run), the command will silently reuse the stale version—even if an updated version exists in your source folder. This can result in building packages with outdated or unintended assets, and creates confusing build results because the target directory is polluted by files from previous invocations.

### To Reproduce

This only occurs when the `package` command is called without an explicit --manifest parameter.

### Expected behavior

The `package` command should never modify or pollute the original target folder. Instead, it should copy all required files to a temporary location, perform packaging operations there, and leave the original target folder and its contents untouched. The build process must always use the most recent source files, and not reuse anything left over from previous packaging runs.

### Screenshots

_No response_

### OS Version and details

_No response_

chiaramooney:
## Description

The `pack` command should never modify or pollute the original input folder. Instead, it should copy all required files to a temporary location, perform packaging operations there, and leave the original target folder and its contents untouched.

## Related Issue

Fixes #281 

## Type of Change

- 🐛 Bug fix
- 🧪 Test update

## Checklist

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -47,7 +47,7 @@ node cli.js help
 
 | Command | When to use |
 |---------|-------------|
-| `init` | First-time project setup. Creates `winapp.yaml`, `appxmanifest.xml` with assets, dev certificate, and downloads SDKs. Combines `manifest generate` + `cert generate`. Interactive by default. |
+| `init` | First-time project setup. Creates `winapp.yaml`, `appxmanifest.xml` with assets, and downloads SDKs. Interactive by default. |
 | `restore` | Reinstall packages from existing `winapp.yaml`. Use after clone or when `.winapp/` is missing. Does not update versions. |
 | `update` | Check for newer SDK versions and update `winapp.yaml`. Also refreshes build tools cache. |
 | `manifest generate` | Create `appxmanifest.xml` standalone, without full init. Use when you only need a manifest. |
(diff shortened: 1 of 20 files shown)


src/winapp-CLI/WinApp.Cli/Services/MsixService.cs
```diff
@@ -1172,6 +1179,25 @@ private static void TryDeleteFile(FileInfo path)
...
+            file.CopyTo(Path.Combine(destination.FullName, file.Name), overwrite: true);
+        }
+
+        foreach (var subDir in source.EnumerateDirectories())
+        {
+            var destSubDir = new DirectoryInfo(Path.Combine(destination.FullName, subDir.Name));
+            CopyDirectoryRecursive(subDir, destSubDir);
+        }
+    }
+
```
azchohfi, line 1200: Just one small feedback. This is recursive, and node_modules can get quite deep, so I would avoid doing it this way. You could generalize what we already have in `PrepareRuntimeForPackagingAsync` and use that. It enumerates all files in all directories, so we don't need recursion. I also just checked, and we are using GetFiles, which is not ideal, so just replace it with EnumerateFiles and we should be good.

PR #295, opened 2026-02-18
Added short description for all commands

nmetulev:
## Description

In previous PR, we updated the description to all commands to be much longer for better LLM integration. However, that caused the help command to be overley descriptive and not easily scannable. This PR adds a short description for the top level commands and updates how the `--help` command renders to improve rendering. 

## Type of Change

- ♻️ Refactoring

## Checklist

## Screenshots / Demo
Before:
image

After:
image

## AI Description

This update introduces short descriptions for all top-level commands to enhance the readability of the help command output. The changes include adjustments to the command descriptions and updates to the `--help` command rendering. As a result, commands can now be more easily scanned by users seeking information quickly. 

Example usage:
```bash
winapp --help
```

### DIFF
--- a/.github/workflows/pr-description.yml
+++ b/.github/workflows/pr-description.yml
@@ -84,12 +84,23 @@ jobs:
 
             1. **AI_DESCRIPTION**: A concise description of the PR changes for release notes automation. Include:
                - What changed and why (1-3 sentences, user-facing language)
-               - Code usage examples if new commands, flags, or APIs were added (as fenced code blocks)
-               - A "**Breaking Change:**" callout if existing behavior changed, APIs were removed/renamed, defaults changed, or command signatures changed. Describe what breaks and how to migrate.
+               - Code usage examples if new public commands, flags, or APIs were added (as fenced code blocks)
+               - A "**Breaking Change:**" callout ONLY if the change would break existing users who upgrade without modifying their code or configuration. Specifically, flag as breaking only if:
+                 • A previously existing command, flag, or API was removed or renamed
+                 • The default value of an existing option was changed in a way that alters previous behavior
+                 • An existing command's output format changed in a way that breaks automation
+                 • A required argument was added to an existing command (not a new command)
+               - Do NOT flag as breaking change:
+                 • Adding new commands, subcommands, or flags
+                 • Adding new optional parameters to existing commands
+                 • Adding new features that don't affect existing usage
+                 • Internal refactors that don't change public CLI surface
+                 • Changes to build scripts, tests, CI, or docs
+               - When uncertain whether something is a breaking change, do NOT flag it as breaking.
 
             2. **SUGGESTED_LABELS**: A comma-separated list of labels to apply. Only use labels from this exact set: ${ALLOWED_LABELS.join(', ')}
                - Always suggest exactly one type label (bug, enhancement, documentation)
-               - Add "breaking-change" if there are breaking changes
+               - Add "breaking-change" ONLY if you are confident there is a genuine breaking change per the criteria above. When in doubt, omit it.
                - Add framework labels (electron, dotnet, rust, python) if the change is framework-specific
 
             Format your response exactly like this:


src/winapp-CLI/WinApp.Cli/Commands/IShortDescription.cs
```diff
@@ -0,0 +1,14 @@
...
+// Licensed under the MIT License.
+
+namespace WinApp.Cli.Commands;
+
+/// <summary>
+/// Provides a short description for a command, used in the help listing.
+/// The long description (set via the Command constructor) remains used for
+/// --cli-schema, LLM documentation, and individual command --help pages.
+/// </summary>
+internal interface IShortDescription
```
azchohfi, line 11: Make TCommand be Comand and IShortDescription, then we get compilation checks as well as the test.
Change [here](https://github.com/microsoft/winappCli/blob/f0831f1656283cc73cf9de20b070a0097baefdaa/src/winapp-CLI/WinApp.Cli/Helpers/HostBuilderExtensions.cs#L66) and [here](https://github.com/microsoft/winappCli/blob/f0831f1656283cc73cf9de20b070a0097baefdaa/src/winapp-CLI/WinApp.Cli/Helpers/HostBuilderExtensions.cs#L85) to `where TCommand : Command, IShortDescription`

src/winapp-CLI/WinApp.Cli/Helpers/CustomHelpAction.cs
```diff
@@ -0,0 +1,209 @@
...
+    /// </summary>
+    internal IReadOnlyCollection<string> CategorizedCommandNames =>
+        _categories.SelectMany(c => c.CommandNames).ToArray();
+
+    /// <summary>
+    /// Creates a new custom help action with the specified command categories.
+    /// </summary>
+    /// <param name="targetCommand">The command this custom help applies to (e.g. root command).</param>
+    /// <param name="categories">Ordered list of (category name, command names) tuples.</param>
+    public CustomHelpAction(Command targetCommand, params (string Category, string[] CommandNames)[] categories)
```
azchohfi: We could accept the Command subclasses/types of the commands we want to show in the help, instead of the array of strings, then its strongly typed and we won't forget to update them whenever they change.

src/winapp-CLI/WinApp.Cli/Helpers/CustomHelpAction.cs
```diff
@@ -0,0 +1,209 @@
...
+
+        // Create a Spectre console that writes to the invocation output writer
+        var ansiSupport = Console.IsOutputRedirected ? AnsiSupport.No : AnsiSupport.Detect;
+        var terminalWidth = Math.Max(40, AnsiConsole.Profile.Width);
+        var spectreConsole = AnsiConsole.Create(new AnsiConsoleSettings
+        {
+            Out = new AnsiConsoleOutput(output),
+            Ansi = ansiSupport
+        });
+        spectreConsole.Profile.Width = terminalWidth;
```
azchohfi: Not sure why this is being done this way... You could just use `AnsiConsole` (static) here, but I'm pretty sure this is not mockable in our tests. Since we already have an `IAnsiConsole` instance that is in the dependency service, I would rather not create another one here. Just inject it in WinAppRootCommand's ctor, and pass it down to this class' ctor.

src/winapp-CLI/WinApp.Cli/Helpers/CustomHelpAction.cs
```diff
@@ -0,0 +1,209 @@
...
+    {
+        // Account for the 1-space indent so content fits within the terminal width
+        var width = Math.Max(40, targetConsole.Profile.Width - 1);
+        using var writer = new StringWriter();
+        var renderConsole = AnsiConsole.Create(new AnsiConsoleSettings
+        {
+            Out = new AnsiConsoleOutput(writer),
+            Ansi = targetConsole.Profile.Capabilities.Ansi ? AnsiSupport.Yes : AnsiSupport.No
+        });
+        renderConsole.Profile.Width = width;
```
azchohfi: Same here. I'm pretty sure we don't need to create multiple AnsiConsoles.

PR #302, opened 2026-02-20
Auto generating npm commands and docs

nmetulev:
## Description

This PR exposes all winapp CLI commands as typed JS/TS functions in the npm package. It adds a script to autogenerate those exports from the cli-schema so they are always up to date. It also adds a script to auto generate the npm documentation.

## Usage Example

```typescript
import { init, packageApp, certGenerate } from '@microsoft/winappcli';

// Initialize a new project with defaults
await init({ useDefaults: true });

// Generate a dev certificate
await certGenerate({ install: true });

// Package the built app
await packageApp({ inputFolder: './dist', cert: './devcert.pfx' });
```

## Type of Change

- ✨ New feature
- 📝 Documentation
- 🔧 Config/build

## Checklist

- [x] Main [README.md](../README.md) updated (if applicable)

## AI Description

This PR introduces the auto-generation of JavaScript/TypeScript functions for all winapp CLI commands, ensuring they remain up-to-date with the CLI schema. It includes scripts for both generating exports and npm documentation, allowing users to easily import and use commands in their projects. For example:

```typescript
import { init, packageApp, certGenerate } from '@microsoft/winappcli';

// Initialize a new project with defaults
await init({ useDefaults: true });
```

### DIFF
--- a/scripts/build-cli.ps1
+++ b/scripts/build-cli.ps1
@@ -83,6 +83,28 @@ try
         exit 1
     }
 
+    # Step 1b: Build Node CLI so E2E tests that invoke node cli.js can run
+    if (-not $SkipNpm) {
+        Write-Host "[BUILD] Building Node CLI (for tests)..." -ForegroundColor Blue
+        Push-Location (Join-Path $ProjectRoot "src\winapp-npm")
+        try {
+            npm ci --ignore-scripts
+            if ($LASTEXITCODE -ne 0) {
+                Write-Warning "npm ci failed, Node E2E tests will be skipped"
+            } else {
+                npm run generate-commands
+                npm run compile
+                if ($LASTEXITCODE -ne 0) {
+                    Write-Warning "Node CLI compile failed, Node E2E tests will be skipped"
+                } else {
+                    Write-Host "[BUILD] Node CLI built successfully" -ForegroundColor Green
+                }
+            }
+        } finally {
+            Pop-Location
+        }
+    }
+
     # Step 2: Run tests (unless skipped)
     if (-not $SkipTests) {
         Write-Host "[TEST] Running tests..." -ForegroundColor Blue


src/winapp-npm/src/index.ts
```diff
@@ -13,10 +28,37 @@ export {
...
+  ManifestUpdateAssetsOptions,
+  CertGenerateOptions,
+  CertInstallOptions,
+  PackageOptions,
+  SignOptions,
+  CreateDebugIdentityOptions,
+  GetWinappPathOptions,
+  ToolOptions,
+  StoreOptions,
+} from './winapp-commands';
```
azchohfi: Would there be a way for this to export all *Options without naming each one? Ideally, a new command in the main cli should automatically be exposed by the npm package without any changes in the `.ts` files, aside from running the generate commands, which are now part of the build. Same for the imports at the top of this file too.

src/winapp-npm/src/index.ts
```diff
@@ -30,6 +72,21 @@ export {
...
+  manifestGenerate,
+  manifestUpdateAssets,
+  certGenerate,
+  certInstall,
+  packageApp,
+  sign,
+  createDebugIdentity,
+  getWinappPath,
+  tool,
+  store,
```
azchohfi: Same here.

src/winapp-npm/src/index.ts
```diff
@@ -40,4 +97,17 @@ export default {
...
```
azchohfi: And here.

src/winapp-npm/scripts/generate-commands.mjs
```diff
@@ -0,0 +1,413 @@
...
+const KNOWN_UNIONS = {
+  'WinApp.Cli.Models.IfExists': {
+    tsName: 'IfExistsPolicy',
+    values: ['error', 'skip', 'overwrite'],
+  },
+  'WinApp.Cli.Models.ManifestTemplates': {
+    tsName: 'ManifestTemplate',
+    values: ['packaged', 'sparse'],
+  },
+};
```
azchohfi: Do we need these if we are generating the inline unions a few lines down from here?

src/winapp-npm/package.json
```diff
@@ -8,14 +8,17 @@
...
+    "generate-docs": "node scripts/generate-docs.mjs",
+    "generate-docs:check": "node scripts/generate-docs.mjs --check",
     "compile": "tsc",
     "compile:watch": "tsc --watch",
     "lint": "eslint src/",
     "lint:fix": "eslint src/ --fix",
     "format": "prettier --write src/",
     "format:check": "prettier --check src/",
-    "build": "npm run format:check && npm run lint && npm run compile && npm run build-cli",
+    "build": "npm run generate-commands && npm run format:check && npm run lint && npm run compile && npm run build-cli",
```
azchohfi, line 21: Shouldn't we run "generate-docs" here instead of in the `build-cli.ps1`?

azchohfi (commented):
Awesome to see this automated!

src/winapp-npm/scripts/generate-commands.mjs
nmetulev: I'll update the cli to emit these as well

src/winapp-npm/package.json
nmetulev: Let's keep it in the build-cli for now since it modifies docs outside of the npm package

azchohfi (approved):

PR #320, opened 2026-02-25
Add Assets to resources.pri + Support Qualified Names in Appxmanifest

### LINKED ISSUE #298: [Bug]: `winapp manifest update-assets` only produces one unplated asset
### Describe the bug

The `winapp manifest update-assets` generates only one unplated asset: Square44x44Logo.targetsize-24_altform-unplated.png. 

In Electron Gallery, when I use the generated assets and updated manifest as is. My app's icon in the task bar has a blue background behind it: 

Image

I believe this might be because the asset the taskbar is using is not unplated. Would it be possible that the cli hasn't generated the right dimensioned unplated asset for my monitor size/scale? 

### To Reproduce

1. Create an electron app
2. Create assets for the electron app using winappcli
3. Package the electron app into an msix and run
4. Observe icon in taskbar. 

### Expected behavior

Icon should have no background. 

### Screenshots

Image

### OS Version and details

_No response_

chiaramooney:
Closes #298 

## Description

- Adding assets to resources.pri to allow for unplated assets to be used when an app appears in the Taskbar. resources.pri now includes mentions of all assets mentioned in appxmanifest.
- CopyAllAssetsAsync and PRI generation both required the asset paths in appxmanifest to be enumerated. Extracted that code from CopyAllAssetsAsync so code can be shared for both functions and only run once per pack command.  
- Removes blue background from app icon in the taskbar.  
- When scaled asset was specified in appxmanifest, the other scaled assets of the same name were not included in the Assets folder during packaging. Now if Logo.scaled-100.png is specified in appxmanifest, Logo.scaled-200.png, etc is added to the Assets folder during packaging. The also works for unplated, light/dark, etc variations. 
- **Note:** If a scaled asset like Logo.scaled-100.png is specified in appxmanifest, MRT will not substitute in the right scaled, unplated, etc asset at run time. It will always use the specified asset. For MRT to sub in the most relevant scaled, unplated, etc asset, the developer must specify the base version of the asset (i.e. Logo.png) in the appxmanifest.
- Replaces using Regex-based XML parsing for XmlDocument parser (better for edge cases/more consistent with poor formatted xml)

## Usage Example

No commands change. 

## Type of Change

- 🐛 Bug fix

## Checklist

## Screenshots / Demo

PRI XAML Example: 
```
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<PriInfo>
	<ResourceMap name="electron-gallery" version="1.0" primary="true">
		<Qualifiers>
			<AlternateForm>LIGHTUNPLATED,UNPLATED</AlternateForm>
			<Scale>100</Scale>
			<TargetSize>256,16,20,24,30,32,36,40,48,56,60,64,72,80,96</TargetSize>
		</Qualifiers>
		<ResourceMapSubtree name="Files">
			<ResourceMapSubtree name="Assets">
				<NamedResource name="AppList.png" uri="ms-resource://electron-gallery/Files/Assets/AppList.png">
					<Candidate qualifiers="TargetSize-96, AlternateForm-UNPLATED" type="Path">
						<Value>Assets\AppList.targetsize-96_altform-unplated.png</Value>
					</Candidate>
					<Candidate qualifiers="TargetSize-96, AlternateForm-LIGHTUNPLATED" type="Path">
						<Value>Assets\AppList.targetsize-96_altform-lightunplated.png</Value>
					</Candidate>
...
// More asset information and closing tags
```

Before this change the pri XAML was very simply. Its main entry was just `<ResourceMap name="electron-gallery" version="1.0" primary="true">`.

## AI Description

This pull request enhances the resource management in the WinApp CLI by adding support for including various asset types in the resources.pri file, enabling unplated assets to appear correctly in the Taskbar. Additionally, it refactors asset handling in the appxmanifest to ensure that scaled assets are included in the packaging process when specified. The XML parsing approach has also been improved for better reliability.


src/winapp-CLI/WinApp.Cli/Services/MsixService.cs
```diff
@@ -514,7 +514,18 @@ public async Task<FileInfo> CreatePriConfigAsync(DirectoryInfo packageDir, TaskC
         }
 
         var resfilesPath = Path.Combine(packageDir.FullName, "pri.resfiles");
-        var priFiles = (packageDir.EnumerateFiles("*.pri").Select(di => di.FullName)).ToList();
+        var assetsDir = new DirectoryInfo(Path.Combine(packageDir.FullName, "Assets"));
```
nmetulev: This works if all the assets are in the "Assets" folder, but it wouldn't work if the assets are in a different location. We should instead check the manifest and locate all referenced assets there, and index them

src/winapp-CLI/WinApp.Cli/Services/MsixService.cs
```diff
@@ -556,6 +616,295 @@ public async Task<FileInfo> CreatePriConfigAsync(DirectoryInfo packageDir, TaskC
...
+        ".png",
+        ".jpg",
+        ".jpeg",
+        ".gif",
+        ".bmp",
+        ".ico",
+        ".svg"
+    };
+
+    private static async Task<HashSet<string>> GetManifestReferencedFilePathsAsync(FileInfo manifestPath, CancellationToken cancellationToken)
```
nmetulev: Could we reuse the `ManifestService.ExtractAssetRefrencesFromManifest` function here somehow instead or doing this work - they do a similar job, no?

PR #420, opened 2026-04-09
Merge VS Code Extension into main

chiaramooney:
## Description

Merge VS Code extension into main. Includes Winapp commands, launch and debug, and README. Does not include extension in publish. 

Adds WinUI templates.

## Usage Example

Install extension in VS Code and use commands/launch windows apps. 

## Type of Change

- ✨ New feature

## Checklist

- [ ] New tests added for new functionality (if applicable)
- [ ] [docs/usage.md](../docs/usage.md) updated (if CLI commands changed)

## AI Description

This pull request merges the VS Code extension into the main codebase, introducing Winapp commands, launch, and debug features. Additionally, it adds WinUI templates and includes updated documentation reflecting these changes. Users can now install the extension and utilize commands for launching Windows apps directly from VS Code. 

```bash
# Add alias inferred from the Executable attribute in the manifest
winapp manifest add-alias

# Specify the alias name explicitly
winapp manifest add-alias --name myapp

# Target a specific manifest file
winapp manifest add-alias --manifest ./path/to/appxmanifest.xml
```


chiaramooney:
@copilot apply changes based on the comments in [this thread](https://github.com/microsoft/winappCli/pull/420#pullrequestreview-4092778803)

docs/getting-started.md
```diff
@@ -0,0 +1,157 @@
+# Getting Started with WinApp CLI
```
nmetulev: should this guide be named more specific - is it about getting started with the vscode extension and packaged apps? and should it be part of the guides?

chiaramooney: @nmetulev Alex wrote the doc but it looks like the guide covers both winapp cli + the extension getting started info. It isn't quite a set by step walk through of one app like the guides are. But focuses more on different scenarios you can use winapp cli tooling for. And is different than the usage guide which lists all the api's and their descriptions. Looks like he wrote the guide end of Feb.

Looks like we have a electron getting started in the same folder, so this would be the general equivalent: 
image

But some of the content is odd, like the installation instructions talk about running a powershell script

chiaramooney: After reading through I think we should delete this guide. And just refer people to use the platform specfic-guides. And then I'm going to create a stand-along guide for getting started using the VS Code extension. I'll do that in a separate PR after this one is merged.

PR #423, opened 2026-04-13
Init and restore commands use current working directory as default when no parameter is passed

### LINKED ISSUE #421: [Bug]: winapp init requires a base directory when building the sample cpp app.
### Describe the bug

Title

### To Reproduce

This occurs in both the cpp sample and the electron.

1. Create a minimal C++ project
2. This should work but fails: `winapp init`
 The error from winapp init was:
 Unrecognized command or argument 'C:\Users\jaylynbarbee\AppData\Local\Microsoft\WindowsApps\winapp.exe'
 
This works:
 `winapp init .`

### Expected behavior

winapp init should use the directory it is currently in if no path is passed.

### Screenshots

_No response_

### OS Version and details

_No response_

zateutsch:
Title, easy one

fixes #421

### DIFF
--- a/src/winapp-CLI/WinApp.Cli.Tests/InitCommandTests.cs
+++ b/src/winapp-CLI/WinApp.Cli.Tests/InitCommandTests.cs
@@ -19,6 +19,22 @@ public class InitCommandTests : BaseCommandTests
     }
 
     [TestMethod]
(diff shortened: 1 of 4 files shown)


zateutsch:
@copilot apply changes based on the comments in [this thread](https://github.com/microsoft/winappCli/pull/423#pullrequestreview-4102245398)

PR #424, opened 2026-04-13
Add daily update check and upgrade command to handle upgrading from CLI

### LINKED ISSUE #392: [Feature]: CLI should check for updates

The CLI should let the user known when a new update is available and show them how to update to latest based on how it was installed (winget, exe, npm, etc)

The cli should check daily if a new release is available that has a higher version that the current version. if there is, it should print with "v0.x.x is available, run <command> to update". This should only happen at most once a day.

A `winapp upgrade` command should handle updating the cli based on the install channel:
- if installed as msix, it would download the latest msix and install it
- if just an exe, it would download the new exe and swap in place
- otherwise, if installed via npm or nuget, it would ask the user to use their package manager to update

### Additional context

_No response_

zateutsch:
fixes #392 

This PR adds a daily check for a new version of WinAppCli and adds an `upgrade` command that handles upgrading the user's CLI to the latest version with whatever original installation method they used.


zateutsch:
@copilot apply changes based on the comments in [this thread](https://github.com/microsoft/winappCli/pull/424#pullrequestreview-4102580691)

zateutsch:
@copilot apply changes based on the comments in [this thread](https://github.com/microsoft/winappCli/pull/424#pullrequestreview-4102580691)

src/winapp-CLI/WinApp.Cli/Services/CliUpgradeService.cs
```diff
@@ -0,0 +1,428 @@
...
+
+    public async Task<int> UpgradeAsync(CancellationToken cancellationToken = default)
+    {
+        var channel = DetectInstallChannel();
+
+        switch (channel)
+        {
+            case InstallChannel.Npm:
+                logger.LogInformation("winapp was installed via npm. To upgrade, run:");
+                logger.LogInformation("  npm update -g @microsoft/winappcli");
```
nmetulev: we should make this more generic in case they use a different package manager like yarn

src/winapp-npm/package.json
```diff
@@ -1,75 +1,75 @@
 {
-  "name": "@microsoft/winappcli",
```
nmetulev, line 2: this file should not be modified, please revert

src/winapp-CLI/WinApp.Cli/Services/CliUpgradeService.cs
```diff
@@ -0,0 +1,428 @@
...
+    {
+        var channel = DetectInstallChannel();
+        var upgradeHint = channel switch
+        {
+            InstallChannel.Npm => "npm update -g @microsoft/winappcli",
+            InstallChannel.NuGet => "update the Microsoft.Windows.SDK.BuildTools.WinApp NuGet package",
+            _ => "winapp upgrade"
+        };
+
+        AnsiConsole.MarkupLine($"[yellow]v{newVersion} is available. Run `{Markup.Escape(upgradeHint)}` to update.[/]");
```
nmetulev, line 335: This reads awkwardly for the nuget package (Run update the bla Nuget package).

nmetulev (changes requested):
Can you update the description of the pr - we use the description for release notes and blog posts and the more info we have the better. 

Looks like the upgrade command needs to be added to the generate-llm-docs.ps1 script so it's added to the skills

src/winapp-CLI/WinApp.Cli/Services/CliUpgradeService.cs
zateutsch:
```suggestion
            case InstallChannel.Npm:
                logger.LogInformation("winapp was installed via Node. Upgrade with your Node package manager.");
```

src/winapp-CLI/WinApp.Cli/Services/CliUpgradeService.cs
```diff
@@ -0,0 +1,427 @@
...
+        return (downloadUrl, version);
+    }
+
+    private void DisplayUpdateNotification(string newVersion)
+    {
+        var channel = DetectInstallChannel();
+        var upgradeHint = channel switch
+        {
+            InstallChannel.Npm => "npm update -g @microsoft/winappcli",
+            InstallChannel.NuGet => "update the Microsoft.Windows.SDK.BuildTools.WinApp NuGet package",
```
zateutsch:
```suggestion
            InstallChannel.NuGet => "dotnet add package Microsoft.Windows.SDK.BuildTools.WinApp",
```

nmetulev:
Pushed 03c5835 to fix the build break. The recent merge of main dropped the declarations of \
ootCommand\ and \parseResult\ in Program.cs while leaving 5 call sites that referenced them. Restored the canonical pattern from main (single \
ootCommand\ + \parseResult\, including the \WINAPP_CLI_CALLER\ env var setup) and removed the leftover \ffectiveParseResult ??\ guard that had no nullable input. Local build clean; full test suite 787/787 passing.

nmetulev (approved):

nmetulev (changes requested):
Did an end-to-end review (code + manual smoke test on a Release-published build). Full test suite is green (787/787). Findings below — one is a blocker, a few others worth addressing before merge.

## 🔴 Critical — must fix before merge

**1. `winapp upgrade` can silently DOWNGRADE users.** `UpgradeAsync` (CliUpgradeService.cs:136) never calls `IsNewerVersion` — it unconditionally downloads and installs whatever `/releases/latest` returns. I reproduced this: published a fresh Release build (v1.0.0+`<sha>`, 23 MB), ran `winapp upgrade`, and my binary was replaced with v0.2.1 (18 MB), leaving a `.old` backup. Anyone on a nightly/dev/preview build running `winapp upgrade` will be silently downgraded. Suggested fix: gate `UpgradeMsixAsync`/`UpgradeExeAsync` on `IsNewerVersion(latest, current)` and print "already up to date" otherwise, and/or add a `--force` flag.

**2. No signature/checksum verification** on the downloaded MSIX/EXE (lines ~180 and ~230). The exe rename trick then puts an unverified binary in place. Given this is an auto-update path, at minimum the downloaded artifact should have its Authenticode signature verified before install.

## 🟡 Should-fix

**3. No overall timeout on `UpgradeAsync`.** The static `HttpClient` has no `Timeout`; only `CheckAndNotifyAsync` is gated by the 2s `WaitAsync` in `Program.cs`. A slow/hung download from `upgrade` will hang indefinitely.

**4. Notification uses static `AnsiConsole.MarkupLine`** (line 335) instead of the DI'd `IAnsiConsole` — same testability issue I just fixed in #485. No test actually asserts the notification text is shown.

**5. No tests for the risky paths** (MSIX install, standalone exe swap, GitHub API parsing, `IsNewerVersion` edge cases). Tests today cover npm/nuget (which just log) + channel detection + cache behavior.

## 🟢 Minor

**6. Cache file's second line is dead data** — `WriteCacheFile` writes `{timestamp}\n{newVersion}`, but `CheckAndNotifyAsync` reads only `lines[0]`. Either use the version (e.g., to avoid re-notifying the same version next day) or drop it.

**7. `UpgradeCommand_HasCorrectDescription`** asserts on exact description text → brittle.

**8. Update check fires on `winapp --version`**, potentially adding up to 2 s to a query that should be instant.

## ✅ Verified working
- Channel detection (npm/nuget/standalone via `WINAPP_CLI_CALLER`)
- 24-hour cache debounce
- `--quiet` / `--json` / `--cli-schema` correctly skip the check
- Notification renders correctly ("v99.0.0 is available...")
- Exe rename self-update actually works on Windows
- PR #485's clean stdout for `get-winapp-path` still holds
- Full suite: **787/787 passing**

Happy to approve once #1 is addressed — that's an active regression risk. #2–#5 are worth doing before merge; the rest are follow-up candidates.

zateutsch:
@copilot apply changes based on the comments in [this thread](https://github.com/microsoft/winappCli/pull/424#pullrequestreview-4151247605)

src/winapp-CLI/WinApp.Cli/Services/CliUpgradeService.cs
```diff
@@ -0,0 +1,557 @@
...
+
+            var dashIdx = value.IndexOf('-');
+            if (dashIdx >= 0)
+            {
+                prerelease = value[(dashIdx + 1)..];
+                value = value[..dashIdx];
+            }
+
+            return Version.TryParse(value, out coreVersion);
+        }
```
zateutsch, line 398:
```suggestion
        static bool TryParseSemVer(string value, out Version coreVersion, out string? prerelease)
        {
            prerelease = null;
        
            var plusIdx = value.IndexOf('+');
            if (plusIdx >= 0)
                value = value[..plusIdx];
        
            var dashIdx = value.IndexOf('-');
            if (dashIdx >= 0)
            {
                prerelease = value[(dashIdx + 1)..];
                value = value[..dashIdx];
            }
        
            // Parse into a nullable to satisfy nullable analysis, then assign on success
            if (Version.TryParse(value, out Version? parsed) && parsed is not null)
            {
                coreVersion = parsed;
                return true;
            }
        
            coreVersion = new Version(0, 0); // any non-null default; will be ignored because we return false
            return false;
        }
```

src/winapp-CLI/WinApp.Cli.Tests/CliUpgradeServiceTests.cs
```diff
@@ -0,0 +1,860 @@
...
+    {
+        // "2.0.0-rc.1+sha.abc" stripped to "2.0.0" > "1.0.0"
+        Assert.IsTrue(CliUpgradeService.IsNewerVersion("2.0.0-rc.1+sha.abc", "1.0.0"));
+    }
+
+    [TestMethod]
+    public void IsNewerVersion_CurrentHasPrerelease_LatestIsStable()
+    {
+        // Current "1.0.0-beta" stripped to "1.0.0", same as latest "1.0.0" — not newer
+        Assert.IsFalse(CliUpgradeService.IsNewerVersion("1.0.0", "1.0.0-beta"));
```
zateutsch:
```suggestion
        Assert.IsTrue(CliUpgradeService.IsNewerVersion("1.0.0", "1.0.0-beta"));
```

PR #452, opened 2026-04-16
[CLI] Shell completion for commands, options, and values

### LINKED ISSUE #357: [Feature]: Shell completion

I would like to have shell completion for PowerShell

For example, `winget` has `complete` subcommand:
```pwsh
Register-ArgumentCompleter -Native -CommandName winget -ScriptBlock {
	param($wordToComplete, $commandAst, $cursorPosition)
	winget complete --word "$wordToComplete" --commandline "$commandAst" --position $cursorPosition | ForEach-Object {
		[System.Management.Automation.CompletionResult]::new($_)
	}
}
```

### Additional context

_No response_

Jaylyn-Barbee:
## Description
This PR gives users a command to turn on completion. When turned on users will be able to type `winapp ` and then press `tab` to loop through each command and read a short description before selecting. This makes our tool easier to use and puts less of a burden to remember all the commands on our users. This also works for nested commands ie: `winapp cert ` and options `winapp init --c`.

## Usage Example
You can press tab to cycle through commands and options.
```bash
$ winapp cert **generate**
$ winapp cert **install**
$ winapp cert **info**
```

You can press ctrl+space to open a list of all available commands or options and their descriptions.
```bash
$ winapp init --c
**--cli-schema**   --config-dir   --config-only

Output the complete CLI command structure as JSON for tooling, scripting, and LLM integration. Includes all commands, options, arguments, and their descriptions.  
```

## Related Issue
Fixes: https://github.com/microsoft/winappCli/issues/357

## Type of Change

- ✨ New feature
- 🧪 Test update

## Checklist

- [X] [docs/usage.md](../docs/usage.md) updated (if CLI commands changed)
- [x] [Language-specific guides](../docs/guides) updated (if applicable) --> shell-completion.md added

## Screenshots / Demo
https://github.com/user-attachments/assets/44c756a2-cb90-4575-883c-984098271d40

## Testing Locally
 
 To test this branch, check it out and follow these steps:
 
 ### 1. Build the CLI
 
 ```powershell
 git checkout jay/shell-completion
 dotnet build src/winapp-CLI/WinApp.Cli/WinApp.Cli.csproj -c Debug
```

2. Register the completer (current session only)

```powershell
 Set-PSReadLineKeyHandler -Key Tab -Function MenuComplete
 Register-ArgumentCompleter -Native -CommandName winapp 
-ScriptBlock {
     param($wordToComplete, $commandAst, $cursorPosition)
     if ($wordToComplete -like '.*' -or $wordToComplete -like '\*' -or $wordToComplete -like '[A-Z]:\*') { return }
     if ($wordToComplete -eq '--') {
         [System.Management.Automation.CompletionResult]::new('-- ', '-- ', 'Text', 'End of options')
         return
     }
     $exe = "<REPO_PATH>\src\winapp-CLI\WinApp.Cli\bin\Debug\net10.0-windows10.0.19041.0\win-arm64\winapp.dll"
     dotnet exec $exe complete "--commandline=$commandAst" "--position=$cursorPosition" 2>$null | ForEach-Object {
         $parts = $_ -split "`t", 2
         $label = $parts[0]
         $tip = if ($parts.Count -gt 1) { $parts[1] } else { $label }
         [System.Management.Automation.CompletionResult]::new($label, $label, 'ParameterValue', $tip)
     }
 }
```

 Replace <REPO_PATH> with your local repo path, and win-arm64
 with win-x64 if on an x64 machine.

3. Try it out
winapp /Tab/→ Shows all commands with descriptions
winapp cert /Tab/ → generate, info, install
winapp init /Tab/ → Shows all options (--setup-sdks, --config-dir, etc.)
winapp init --c/Tab/ → --cli-schema, --config-dir, --config-only
winapp i/Tab/ → init
winapp init ./Tab/ → Falls through to file completion

## Additional Notes
Activates via a command `winapp complete --setup powershell >> $PROFILE` which is listed in the documentation

## AI Description

This PR introduces shell completion for the `winapp` command-line tool, allowing users to easily access commands, subcommands, and options through tab completion. Users can enable this feature in PowerShell with the command `winapp complete --setup powershell`, after which pressing Tab will cycle through commands and options, enhancing the usability of the CLI tool.
```powershell
# Setup for PowerShell (permanent)
winapp complete --setup powershell >> $PROFILE

# Try it in the current session only
winapp complete --setup powershell | Out-String | Invoke-Expression
```


nmetulev (commented):
I initially liked the showing all the commands below when I type, but it looks like that is changing the completion behavior for all commands, not just for winapp - do we want that? Or more importantly, would users want that. Looking at how dotnet does it, it looks like it doesn't change how autocomplete in powershell looks like, and works just like it worked before, but it also knows dotnet commands https://learn.microsoft.com/en-us/dotnet/core/tools/enable-tab-autocomplete

docs/guides/shell-completion.md
```diff
@@ -0,0 +1,67 @@
...
+Registers a native argument completer that calls `winapp complete` on each Tab press, providing context-aware suggestions for commands, subcommands, options, and values.
+
+### To deactivate
+
+Open your profile (`notepad $PROFILE`) and remove the `Register-ArgumentCompleter` block for `winapp`. Restart PowerShell.
+
+## Bash
+
+```bash
+winapp complete --setup bash >> ~/.bashrc
```
nmetulev, line 40: why bash - do we think users will run it in bash

nmetulev (changes requested):
The testing approach is right — BaseCommandTests + TestAnsiConsole.Output + ParseAndInvokeWithCaptureAsync is
consistent with the rest of the repo and exercises the command end-to-end through System.CommandLine. The issue is 
coverage: the 14 tests hit happy paths, but most of the interesting logic in Invoke() is untested, and that's
exactly where regressions will hide.

Coverage gaps (priority order)

1. None of the cursor-position edge cases are tested, even though the production code has nontrivial branches for
them:

 - position > commandLine.Length → clamp + append trailing space (the trickiest line in the file)
 - position < 0 → clamped to 0

Every existing test hardcodes --position to exactly the string length, which is the only case where the clamp logic
is a no-op.

2. The "interesting" filtering branches are untested:

 - End-of-options --  short-circuit — no test.
 - Path fallthrough (currentWord.StartsWith('.') / '\\'). The PR description explicitly calls this out as a feature 
("winapp init ./Tab/ → Falls through to file completion") but there's no test.
 - Alias exclusion — aliasNames filtering, including the explicit -h/-?//h//? adds. No test asserts these are absent
 from completions.
 - Commands-vs-flags preference + fallback — when no subcommands exist (e.g. winapp init ) the code falls back to 
showing flags. Neither branch is tested.

3. Prefix vs substring matching is barely tested. Complete_PartialCommand_ReturnsMatchingCommands only asserts init
is present — it doesn't assert that cert/package/etc. are absent. So it would still pass with the default
System.CommandLine substring matching that this PR explicitly overrides.

4. The headline example from the PR description isn't tested. winapp init --c → --cli-schema, --config-dir,
--config-only is the canonical demo but has no test.

5. Empty/no-commandline tests only check exit code, not that nothing was written. The function returns 0 early, so
the assertion should also be Assert.AreEqual(string.Empty, GetOutput()) to catch accidental output.

Other issues

6. CliSchema_DoesNotContainHiddenCommands is in the wrong file — it tests CliSchema behavior, not CompleteCommand.
Either move it to a CliSchemaTests file or add a comment explaining why it lives here.

7. Setup-script tests are surface-level grep checks. Acceptable as smoke tests, but consider asserting bash/zsh
scripts also reference --commandline and --position (only the powershell test does that today). And 
Complete_SetupUnknownShell_ReturnsError should verify something was written to stderr — a silent exit-1 would still
pass.

8. Mojibake in the test file. Several comments contain â€" instead of — (em-dash). Save the file as UTF-8 (with BOM
if MSBuild requires it) or replace the em-dashes with --.

Suggested additions

 [TestMethod] Complete_PartialOption_PrefixMatches              // "winapp init --c" → only --cli-schema / 
--config-*
 [TestMethod] Complete_PartialCommand_ExcludesNonMatching       // assert "cert" NOT in results for "winapp in"
 [TestMethod] Complete_PathPrefix_ReturnsEmpty                  // "winapp init ./" → no completions
 [TestMethod] Complete_EndOfOptionsMarker_ReturnsEmpty          // "winapp init -- something " → no completions
 [TestMethod] Complete_PositionPastEnd_TreatsAsTrailingSpace    // e.g. position=20 on 11-char string
 [TestMethod] Complete_NegativePosition_ClampsToZero
 [TestMethod] Complete_AliasesAreExcluded                       // -h, -?, and any real subcommand alias
 [TestMethod] Complete_EmptyCommandLine_ProducesNoOutput        // strengthen existing test
 [TestMethod] Complete_NoSubcommands_FallsBackToOptions         // "winapp init " shows --options

Bottom line

Right approach, wrong coverage. As written, the tests would pass even if you regressed the prefix matching, the path
fallthrough, the -- short-circuit, the alias filter, or the position clamping — i.e. all the custom logic this PR
layers on top of System.CommandLine. Adding the ~9 tests above would close the gap.

docs/guides/shell-completion.md
Jaylyn-Barbee: Because we used existing completion models for inspiration (ie: dotnet/winget) and those models are cross platform, we ended up with coverage for bash/zsh (WSL). So, the docs just included instructions for that coverage.

nmetulev (approved):

PR #459, opened 2026-04-17
[Telemetry] Changes to detect if an agent is using winapp

Jaylyn-Barbee:
## Description
Adds two new telemetry fields to every event to distinguish whether a CLI invocation comes from an AI coding agent, CI, or a human:
- **SenderOrigin**: "agent", "ci", or "direct"
- **AgentName**: normalized agent name when detected (e.g., "copilot", "claude-code", "cursor"), null otherwise

New `AgentEnvironmentDetector` checks generic standards (AI_AGENT, AGENT) then tool-specific env vars (COPILOT_MODEL, CLAUDECODE, CURSOR_CLI, CODEX_CI, GOOSE_TERMINAL, etc.), falling back to CI detection, then defaulting to "direct". Results are cached per-process.

## Usage Example
The fields are automatically populated on every telemetry event.
To verify locally via PerfView:

Direct invocation
```bash
 PerfView.exe /onlyProviders="*Microsoft.Windows.WinAppDevCLI" collect
 winapp --help
 # Stop collection → Events → filter "Microsoft.Windows.WinAppDevCLI"
 # Expect: SenderOrigin="direct", AgentName=""
```
 
Simulated agent invocation
```bash
 PerfView.exe /onlyProviders="*Microsoft.Windows.WinAppDevCLI" collect
 $env:AI_AGENT="copilot"; 
 winapp --help
 Remove-Item Env:AI_AGENT
 # Stop collection → Events → filter "Microsoft.Windows.WinAppDevCLI"
 # Expect: SenderOrigin="agent", AgentName="copilot"
```

## Related Issue
- Fixes: "Telemetry for commands being used by agents"

## Type of Change
 - ✨ New feature
 - 📈 Telemetry

## Checklist

Additional Notes
 - Detection covers: GitHub Copilot, Claude Code, Cursor, Codex, Goose, Cline plus the emerging generic AI_AGENT/AGENT standards
 - Adding a new agent in the future = one line in AgentEnvironmentDetector.ToolSpecificAgentVariables
 - The existing CI boolean field is preserved for backward compatibility; `SenderOrigin="ci"` provides the same signal within the unified enum
 - 18 new unit tests


src/winapp-CLI/WinApp.Cli/Telemetry/AgentEnvironmentDetector.cs
```diff
@@ -0,0 +1,149 @@
...
+        ("CURSOR_AGENT", "cursor"),
+        ("CURSOR_CLI", "cursor"),
+
+        // OpenAI Codex CLI - https://github.com/openai/codex
+        ("CODEX_CI", "codex"),
+
+        // Goose (Block) - https://github.com/block/goose
+        ("GOOSE_TERMINAL", "goose"),
+
+        // GitHub Copilot
```
nmetulev, line 57: you are missing few, this is what I see in copilot cli, i think we need at least COPILOT_CLI 

Image

Or we separate cli from vscode

src/winapp-CLI/WinApp.Cli/Telemetry/Telemetry.cs
```diff
@@ -151,6 +151,9 @@ public void LogException(string action, Exception e, Guid? relatedActivityId = n
             innerException = innerException.InnerException;
         }
 
+        var (senderOrigin, agentName) = AgentEnvironmentDetector.Detect();
+        var sanitizedAgentName = agentName != null ? this.ReplaceSensitiveStrings(agentName) : null;
+
         this.LogInternal(
```
nmetulev, line 157: The anonymous-type LogXxx methods are the wrong place to add these fields.
 
 LogException and LogTimeTaken build anonymous types instead of EventBase subclasses, so they never inherited CI, 
 Caller, or AppVersion either. Adding SenderOrigin/AgentName here plugs two of the gaps but leaves the others, which
 deepens an existing inconsistency rather than fixing the root cause.
 
 Recommend converting these two methods to use proper EventBase subclasses (e.g. ExceptionThrownEvent : EventBase, 
 TimeTakenEvent : EventBase). Then all cross-cutting fields — CI, Caller, AppVersion, SenderOrigin, AgentName, 
 PartA_PrivacyProduct — flow through EventBase exactly once, and Telemetry.cs doesn't need to know 
 AgentEnvironmentDetector exists. Bonus: Exception/TimeTaken events start carrying CI/Caller/AppVersion, which is
 probably what we actually want.

src/winapp-CLI/WinApp.Cli/Telemetry/Telemetry.cs
```diff
@@ -216,6 +226,7 @@ public void LogCritical(string eventName, bool isError = false, Guid? relatedAct
     {
         data.ReplaceSensitiveStrings(this.ReplaceSensitiveStrings);
         data.Caller = data.Caller != null ? this.ReplaceSensitiveStrings(data.Caller) : null;
+        data.AgentName = data.AgentName != null ? this.ReplaceSensitiveStrings(data.AgentName) : null;
```
nmetulev, line 229: Sanitization is duplicated across every LogXxx overload.
 
 AgentName sanitization now has to be remembered in Log, LogError, LogException, and LogTimeTaken — same trap that 
 Caller already has. The next LogXxx overload added will silently ship raw AgentName.
 
 Prefer moving this into one place — either fold it into EventBase.ReplaceSensitiveStrings (so subclasses can't
 forget) or have a single private helper in Telemetry that all LogXxx methods call before LogInternal. Same applies
 to Caller while you're in there.

nmetulev (changes requested):

src/winapp-CLI/WinApp.Cli/Telemetry/AgentEnvironmentDetector.cs
Jaylyn-Barbee: Updated:
```csharp
// GitHub Copilot CLI
("COPILOT_CLI", "copilot-cli"),

// GitHub Copilot (VS Code)
("VSCODE_COPILOT_TERMINAL", "copilot-vscode"),

// Generic Copilot fallback
("COPILOT_MODEL", "copilot"),
```

src/winapp-CLI/WinApp.Cli/Telemetry/Telemetry.cs
Jaylyn-Barbee: Updated:  All events now flow through EventBase → Log<T>/LogError<T>, which means CI, Caller, AppVersion, SenderOrigin, AgentName, and PartA_PrivacyProduct are all inherited and sanitized consistently. Telemetry.cs no longer references AgentEnvironmentDetector at all

Jaylyn-Barbee: I did also notice that `LogTimeTaken` and `LogCritical` are never referenced. This is not code I introduced in this PR but I assume we wanted to be logging how long commands and things like that are taking for our users?

src/winapp-CLI/WinApp.Cli/Telemetry/Telemetry.cs
Jaylyn-Barbee: Fixed: moved to one place.

nmetulev (approved):

PR #485, opened 2026-04-20
fix(get-winapp-path): fall back to global cache when local missing (#475)

### LINKED ISSUE #475: [Bug]: Get local Win App path returns path to directory that doesn't exist
### Describe the bug

Run locak version of get win app path:

Returns path `C:\Users\zteutsch\source\tauri-app\.winapp`

This folder does not exist.

### To Reproduce

1. Run `winapp get-winapp-path`
2.  Returns a directory that doesn't exist: `C:\Users\zteutsch\source\tauri-app\.winapp`

### Expected behavior

Should print a warning that there is no local winapp folder, show global location as fallback?

ie:

"There is no .winapp directory local to this directory, the global directory can be found at: <path>"

### Screenshots

_No response_

### OS Version and details

_No response_

nmetulev:
## Summary

Fixes #475.

Previously, winapp get-winapp-path (without `--global`) would print `<cwd>\.winapp` even when no local `.winapp` directory existed anywhere up the directory tree. Build scripts that consumed the output ended up pointing at non-existent folders, since installed packages had actually landed in the global cache.

## Behavior change

When the resolved local `.winapp` directory does not exist, the command now:

1. Emits a warning (`WinappDirectoryService` + `UiSymbols.Warning`) explaining that no local `.winapp` was found.
2. Falls back to the global cache path (`%USERPROFILE%\.winapp` by default).
3. Returns exit code `0` and prints the global path on stdout, so scripts continue to work.

When `--global` is explicitly requested but the global directory is missing, behavior is unchanged: the command still returns a non-zero exit code with an error.

## Repro (before fix)

```
PS> cd (New-Item -ItemType Directory ""C:\Users\nikolame\AppData\Local\Temp\repro-475-$([guid]::NewGuid())"")
PS> winapp get-winapp-path
C:\Users\me\AppData\Local\Temp\repro-475-...\\.winapp     # <-- non-existent
```

## After fix

```
PS> winapp get-winapp-path
WARN  No local .winapp directory found; falling back to the global cache at C:\Users\me\.winapp.
C:\Users\me\.winapp
```

## Implementation notes

- Switched the path emission in `GetWinappPathCommand` from `logger.LogInformation` to `IAnsiConsole.WriteLine` so the path is the only thing on stdout and is cleanly captured by tests via `TestAnsiConsole`. (Non-error log levels in `TextWriterLogger` use the static `AnsiConsole`, which bypasses both the injected writer and the DI `IAnsiConsole`, so they aren't visible to test assertions.)
- Added `GetWinappPathCommandTests` with three cases: local exists, local missing -> global fallback, `--global` missing -> error.

## Validation

- New tests: 3 passing.
- Full suite: 767/767 passing.
- Manual repro: empty temp dir prints warning + global path, `exit 0`.

Closes #475

### DIFF
--- a/src/winapp-CLI/WinApp.Cli/Commands/GetWinappPathCommand.cs
+++ b/src/winapp-CLI/WinApp.Cli/Commands/GetWinappPathCommand.cs
@@ -61,7 +61,12 @@ internal class GetWinappPathCommand : Command, IShortDescription
                     if (!winappDir.Exists)
                     {
                         var globalDir = winappDirectoryService.GetGlobalWinappDirectory();
-                        logger.LogWarning("{UISymbol} No local .winapp directory found; falling back to the global cache at {GlobalDir}.", UiSymbols.Warning, globalDir.FullName);
+                        // Write the warning to stderr so stdout stays script-friendly
+                        // (callers do `path = $(winapp get-winapp-path)`). Going through
+                        // ILogger.LogWarning would route via the static AnsiConsole and
+                        // pollute stdout (TextWriterLogger only sends >= Error to stderr).
+                        parseResult.InvocationConfiguration.Error.WriteLine(
+                            $"{UiSymbols.Warning} No local .winapp directory found; falling back to the global cache at {globalDir.FullName}.");
                         winappDir = globalDir;
                         directoryType = "Global (fallback)";
                     }
--- a/src/winapp-CLI/WinApp.Cli/Commands/GetWinappPathCommand.cs
+++ b/src/winapp-CLI/WinApp.Cli/Commands/GetWinappPathCommand.cs
@@ -52,12 +52,8 @@ internal class GetWinappPathCommand : Command, IShortDescription
                     winappDir = winappDirectoryService.GetLocalWinappDirectory();
                     directoryType = "Local";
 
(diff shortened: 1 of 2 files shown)


src/winapp-CLI/WinApp.Cli/Commands/GetWinappPathCommand.cs
```diff
@@ -50,6 +51,25 @@ public override Task<int> InvokeAsync(ParseResult parseResult, CancellationToken
                     // Get the local .winapp directory
                     winappDir = winappDirectoryService.GetLocalWinappDirectory();
                     directoryType = "Local";
+
+                    // #475: GetLocalWinappDirectory falls back to <baseDir>/.winapp when no
```
chiaramooney: Nit: do you want all of the comment information about what the broken state was before? Could start this comment at "Fall back to the global cache..."

chiaramooney (approved):

src/winapp-CLI/WinApp.Cli/Commands/GetWinappPathCommand.cs
nmetulev: done

PR #496, opened 2026-04-23
Fix winapp run picking createdump.exe; share exe detection with winapp package

nmetulev:
When multiple `.exe` files exist in the build output (e.g. `createdump.exe` alongside the actual app), the identity registration path (`winapp run`) picked the first one alphabetically — often `createdump.exe` — causing the app to crash on launch.

**Root cause:** `MsixService.Identity.cs` had its own exe detection logic using `EnumerateFiles` with `AllDirectories`, bypassing the filtering used by `winapp package`.

**Changes:**
- Filter known .NET runtime tools (`createdump.exe`, `apphost.exe`) via a shared `IsRuntimeToolExecutable()` helper.
- Reuse `ResolveManifestPlaceholders` in the identity path so `winapp run` and `winapp package` share the same exe detection, filtering, and error handling.
- Add `--executable` / `--exe` option to `winapp run` (mirrors `winapp package`) so users can disambiguate when the build output legitimately contains multiple exes.
- Restore an explicit "executable not found in output directory" error in the identity path.
- Add regression tests for the createdump / apphost filtering cases.


zateutsch (approved):
lgtm.

Side thought that occurred to me while looking at this,

Would it be helpful to have some sort of way to add optional project specific config?
I'm imagining a situation where a user has an executable they want to regularly ignore.

nmetulev:
> Would it be helpful to have some sort of way to add optional project specific config? I'm imagining a situation where a user has an executable they want to regularly ignore.

In that case, I'd say a better solution is to just specify the exe in the Package.appxmanifest

PR #497, opened 2026-04-23
Add New Commands to VSC Extension

### LINKED ISSUE #491: [VS Code] Some WinApp CLI commands are missing from the VS Code Extension
Some WinApp CLI commands like `winapp manifest add-alias` and `winapp ui` are missing from the VS Code Extension currently. It may not make sense to expose all of these commands in the extension, but we should make a clear decision on which commands make sense to add support for and which don't. 

● Here's my analysis. The extension currently exposes 15 commands. Here's how the remaining CLI commands break down:

  ✅ Recommend adding

manifest add-alias │ Directly called out in the issue. Common workflow step — adding an AppExecutionAlias is something devs do from the editor while editing manifests.
unregister  │ Natural counterpart to run and create-debug-identity which are already exposed. Devs need a way to clean up  sideloaded packages without leaving VS Code.  
cert info  │ Lightweight and helpful — "is my cert expired?" is a common debugging question. Pairs well with the existing cert generate/install commands.      

🤔 Good to have but lower priority

create-external-catalog  │ Niche (TrustedLaunch sparse packages only), but if someone needs it, they'd want it in the extension. 
store  │ Useful for Store submissions, but it's essentially a pass-through to the Store CLI. Could be a simple terminal command rather than a full UI integration. 

  ❌ Recommend NOT adding
ui * (all subcommands)   │ These are designed for programmatic UI automation/testing — they're tools for AI agents and scripts, not interactive editor commands. They'd be awkward as Command Palette entries.       
node create-addon   │ Very specialized for Electron/Node. The npm CLI (npx winapp) is the natural interface for Node      developers. 
node   add/clear-electron-debug-identity  │ Same — Electron-specific, better served by the npm package.

chiaramooney:
## Description

Some WinApp CLI commands like winapp manifest add-alias and winapp ui are missing from the VS Code Extension currently. It may not make sense to expose all of these commands in the extension, but we should make a clear decision on which commands make sense to add support for and which don't.

Adds:

manifest add-alias │ Directly called out in the issue. Common workflow step — adding an AppExecutionAlias is something devs do from the editor while editing manifests.
unregister │ Natural counterpart to run and create-debug-identity which are already exposed. Devs need a way to clean up sideloaded packages without leaving VS Code.
cert info │ Lightweight and helpful — "is my cert expired?" is a common debugging question. Pairs well with the existing cert generate/install commands.

## Usage Example

Same usage as other commands in VS Code Extension.

## Related Issue

Fixes #491 

## Type of Change

- 🐛 Bug fix
- ✨ New feature
- 💥 Breaking change
- 📝 Documentation
- 🔧 Config/build
- ♻️ Refactoring
- 🧪 Test update

## Checklist

## AI Description

This update adds new commands to the WinApp CLI VS Code Extension, enhancing functionality for developers. The following commands are now available: `winapp.manifestAddAlias` for adding execution aliases to manifests, `winapp.unregister` for unregistering sideloaded packages, and `winapp.certInfo` for checking certificate details. Usage is consistent with existing commands in the extension.

```bash
# To add an execution alias
winapp.manifest add-alias

# To unregister a sideloaded package
winapp.unregister

# To check certificate information
winapp.cert info
```


nmetulev (approved):
Approved conditional on addressing the copilot comments

nmetulev (approved):

PR #503, opened 2026-04-27
App update available notice on first run of the day - no upgrade command

zateutsch:
Once a day, on first command run, the CLI will check if a newer version of the CLI is available and show a notification at the top of the command output if so.


src/winapp-CLI/WinApp.Cli/Services/UpdateNotificationService.cs
```diff
@@ -0,0 +1,225 @@
...
+            return coreCompare > 0;
+        }
+
+        // Same core version: a stable release (no pre-release) is newer than a pre-release
+        if (currentPre != null && latestPre == null)
+        {
+            return true;
+        }
+
+        return false;
```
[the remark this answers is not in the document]
zateutsch: @copilot apply changes based on [this feedback](https://github.com/microsoft/winappCli/pull/503#discussion_r3149061098)

src/winapp-CLI/WinApp.Cli/Services/UpdateNotificationService.cs
```diff
@@ -0,0 +1,263 @@
...
+        catch
+        {
+            // Silent failure — never disrupt the user's command
+        }
+    }
+
+    internal async Task<string?> GetLatestVersionAsync(CancellationToken cancellationToken = default)
+    {
+        // Allow overriding the latest version for testing (skips GitHub API call)
+        var overrideVersion = Environment.GetEnvironmentVariable("WINAPP_CLI_LATEST_VERSION");
```
nmetulev: This ends up in the production - do we need it in production?

nmetulev (changes requested):
Instead of blocking and introducing latency, I wonder if we could instead run the update check in the background and show it the next time the command is ran if there is an update, removing any kind of latency. Following this pattern:

on startup:
   - read cache
   - if cache says "update X available" and not yet shown today:
         - print notice (instant, no network)
         - mark "shown today" in cache
 
   - if cache is older than 24h:
         - Task.Run(() => RefreshCacheAsync())   // fire and forget
         - // do NOT await — let it race the command
 
 on shutdown (best-effort):
   - if background task is complete, no-op
   - if still running, abandon it

In addition:
1. We should have a way for a user to opt out of update checks - maybe a env variable that we document
2. Should also disable update checks during CI

src/winapp-CLI/WinApp.Cli/Services/UpdateNotificationService.cs
zateutsch: No we don't. I'll remove

zateutsch:
@copilot apply changes based on the comments in [this thread](https://github.com/microsoft/winappCli/pull/503#pullrequestreview-4193478716)

src/winapp-CLI/WinApp.Cli/Services/UpdateNotificationService.cs
```diff
@@ -0,0 +1,337 @@
...
+        if (b == null) { return -1; }  // pre-release < stable
+
+        var aIds = a.Split('.');
+        var bIds = b.Split('.');
+        var len = Math.Min(aIds.Length, bIds.Length);
+
+        for (var i = 0; i < len; i++)
+        {
+            var aIsNum = int.TryParse(aIds[i], out var aNum);
+            var bIsNum = int.TryParse(bIds[i], out var bNum);
```
nmetulev, line 234: **[high] Notice goes to stdout — will break scripted commands**

`ansiConsole.MarkupLine(...)` writes to the registered `IAnsiConsole`, which is `AnsiConsole.Console` (default = stdout). The notice fires before command invocation for every non-`--json`/non-`--quiet` run, so it silently corrupts the stdout of commands whose contract is "machine-readable single value" — e.g. `winapp get-winapp-path` (whose source comment explicitly says "just the path for easy consumption by scripts"). It will also prepend an unrelated line to `winapp --version` and `winapp --help`, which users frequently pipe to `grep`/parsers.

This is what `npm`, `gh`, `cargo`, etc. all hit and why they emit upgrade notices to stderr.

**Recommendation:** route the notice to stderr — e.g. construct a separate `IAnsiConsole` bound to `Console.Error` and use it only for this notification:

```csharp
var errConsole = AnsiConsole.Create(new AnsiConsoleSettings { Out = new AnsiConsoleOutput(Console.Error) });
errConsole.MarkupLine($"[yellow]v{Markup.Escape(newVersion)} is available. To update, {Markup.Escape(upgradeHint)}.[/]");
```

That keeps the existing gating logic intact and unblocks every script-consumer use case at the same time.

src/winapp-CLI/WinApp.Cli/Services/UpdateNotificationService.cs
```diff
@@ -0,0 +1,337 @@
...
+            logger.LogDebug(ex, "Failed to check for CLI updates.");
+            return null;
+        }
+    }
+
+    private void DisplayUpdateNotification(string newVersion)
+    {
+        var upgradeHint = DetectInstallChannel() switch
+        {
+            InstallChannel.Npm => "npm update -g @microsoft/winappcli",
```
nmetulev, line 167: **[medium] Fire-and-forget refresh may never finish**

Most CLI commands return in well under a second, but `HttpClient.Timeout` is 10s. The `Task.Run` lambda has no awaiter and no graceful-shutdown hook, so on short invocations the process exits before the HTTP call completes and the cache is never populated. In practice, users who run only short commands may never see an update notice.

**Recommendation:** when the cache is empty / has never been populated, await the refresh in the foreground with a tight bounded timeout (e.g. ~250ms) before falling back to fire-and-forget. That keeps steady-state latency at zero (cache hit path is unchanged) while giving first-run a real chance to populate without noticeably slowing anything down.

src/winapp-CLI/WinApp.Cli.Tests/UpdateNotificationServiceTests.cs
```diff
@@ -0,0 +1,383 @@
...
+        Assert.IsTrue(UpdateNotificationService.IsNewerVersion("1.0.0-alpha", "1.0.0-1"));
+    }
+
+    [TestMethod]
+    public void IsNewerVersion_LongerPreReleaseWithMatchingPrefix_ReturnsTrue()
+    {
+        // "beta.1.2" > "beta.1" because more fields when prefix matches
+        Assert.IsTrue(UpdateNotificationService.IsNewerVersion("1.0.0-beta.1.2", "1.0.0-beta.1"));
+    }
+}
```
nmetulev, line 428: **[medium] `GetLatestVersionAsync` HTTP/JSON path is essentially untested**

The HTTP fetch + `tag_name` extract + leading-`v` strip + error suppression in `GetLatestVersionAsync` has no deterministic coverage. The closest tests (`RefreshCacheAsync_WritesUpdateCheckCacheFile`, `RefreshCacheAsync_PreservesLastShownDate`) both pass even when the network call returns nothing, so a regression in JSON parsing or `v`-prefix handling could ship silently.

**Recommendation:** inject the `HttpClient`/`HttpMessageHandler` (or extract the parser into a static helper that takes a JSON string), then add tests for:
- `{"tag_name":"v1.2.3"}` → `1.2.3`
- `{"tag_name":"1.2.3"}` → `1.2.3` (no strip)
- missing/null/empty `tag_name` → `null`
- non-2xx response → `null` (no throw)
- invalid JSON → `null` (no throw)
- network exception → `null` (no throw)

src/winapp-CLI/WinApp.Cli/Program.cs
```diff
@@ -84,15 +84,36 @@ static async Task<int> Main(string[] args)
...
+
+        if (args.Length > 0)
+        {
+            parseResult = rootCommand.Parse(args, WinAppParserConfiguration.Default);
+
+            // Set WINAPP_CLI_CALLER env var from --caller option so telemetry and update checks can use it
+            var caller = parseResult.GetValue(WinAppRootCommand.CallerOption);
+            if (!string.IsNullOrWhiteSpace(caller))
+            {
+                Environment.SetEnvironmentVariable("WINAPP_CLI_CALLER", caller);
```
nmetulev, line 102: **[medium] Program-level gating has no integration coverage**

The new gating logic added here — skip update notice for `--json` / `--quiet` / `--cli-schema` / `--complete`, plus setting `WINAPP_CLI_CALLER` from `--caller` *before* invocation — isn't exercised by any test. The new `UpdateNotificationServiceTests` drive the service directly with env vars already set, so a regression that pollutes `--json` output or fails to plumb `--caller` through would still pass the suite.

**Recommendation:** add a small set of Program-level tests that seed `.update-check` with a newer version and assert:
- stdout for `--json` contains no notice
- stdout for `--quiet` contains no notice
- `--caller npm` produces the npm hint via the real entrypoint

nmetulev (commented):
PR review notes from a multi-dimensional pass. One high-severity item (stdout pollution), one correctness/UX gap on first-run latency, plus two test-coverage suggestions. Skipping a couple of medium nits I had earlier.

- **H1** — Update notice goes to stdout, will break script-parseable commands (e.g. `get-winapp-path`, `--version`, `--help`). Recommend routing the notice to **stderr**.
- **M1** — Background refresh is fire-and-forget `Task.Run`; short-lived invocations exit before the HTTP call completes, so the cache may never populate. Recommend a tight bounded foreground wait (~250ms) when the cache is empty so the first-ever run still has a chance to populate.
- **M3** — `GetLatestVersionAsync` (HTTP + JSON + `v`-strip + error suppression) has no deterministic test.
- **M4** — Program.cs gating for `--json` / `--quiet` / `--caller` plumbing has no integration coverage.

nmetulev (approved):

PR #513, opened 2026-04-30
Make Microsoft.Windows.SDK.BuildTools.WinApp safe for transitive consumption

nmetulev:
## Summary

Makes `Microsoft.Windows.SDK.BuildTools.WinApp` safe for **transitive consumption** so a library author (e.g., the MAUI Windows wrapper) can take a `PackageReference` on it and have the targets flow through to consumer head apps — without having to ask every consumer to add their own `PackageReference`.

### Targets hardening (`build/Microsoft.Windows.SDK.BuildTools.WinApp.targets`)

Introduces a single computed master gate `_WinAppRunSupportActive` that all gated targets/items now reference instead of repeating long conditions. The gate requires **all** of:

1. `EnableWinAppRunSupport != false` (existing opt-out preserved)
2. `_WinAppEffectiveTargetPlatformIdentifier == Windows` (skips iOS/Android/Mac TFMs in MAUI multi-target)
3. `OutputType != Library` (head apps only — `Exe`, `WinExe`, etc.)
4. `WindowsPackageType != None` (skip unpackaged WinUI apps)
5. A manifest exists — either `WinAppManifestPath` is set and the file exists, **or** auto-detection finds one

`_WinAppEffectiveTargetPlatformIdentifier` prefers `$(TargetPlatformIdentifier)` if explicitly set, falls back to `[MSBuild]::GetTargetPlatformIdentifier($(TargetFramework))`.

The gate is also applied to the `<AppxManifest>` ItemGroup, the design-time `<Compile>` inclusion, and the `<Content Include="appxmanifest.xml">` ItemGroup so they no longer fire in unrelated projects.

### Packaging (`Microsoft.Windows.SDK.BuildTools.WinApp.csproj`)

- Added `<None Include="build\**\*" PackagePath="buildTransitive" />` so the targets are mirrored into `buildTransitive/` (verified in the produced nupkg: 2 files in `build/`, identical 2 in `buildTransitive/`).
- Removed `<DevelopmentDependency>true</DevelopmentDependency>` — this was forcing consumers to default to `PrivateAssets=all`, which blocks transitive flow entirely.

### Docs

- `src/winapp-NuGet/README.md`: split Usage into "Direct consumption (head app)" with `PrivateAssets="all"` recommended, and "Transitive consumption (library author)" with `PrivateAssets="none" IncludeAssets="build;buildTransitive"`.
- `docs/dotnet-run-support.md`: full 5-condition gate explanation in the Detection Logic section.

### Tests (`src/winapp-NuGet/tests/NuGet.Tests.ps1`)

New 14-case Pester suite (12 gate matrix + 2 nupkg layout). Runs `dotnet msbuild -getProperty:_WinAppRunSupportActive` against synthetic csprojs (Library / Exe / WinExe × manifest yes/no × `EnableWinAppRunSupport=false` × non-Windows TFM × `WindowsPackageType=None` × custom manifest path missing). All 14 pass.

## Validation

End-to-end smoke tests against the locally produced `0.3.1-nm-winapp-nuget-buildtransitive.10.nupkg`:

| Scenario | OutputType | Manifest | Consumption | Gate | Build |
|---|---|---|---|---|---|
| `samples/dotnet-app` | `Exe` | yes | direct | `true` ✅ | ✅ |
| `samples/wpf-app` | `WinExe` (UseWPF) | yes | direct | `true` ✅ | ✅ |
| Synthetic `MyLib` | `Library` | no | direct (PrivateAssets=none) | `false` ✅ | ✅ |
| Synthetic `HeadExe` (no manifest) | `Exe` | no | transitive via ProjectReference | `false` ✅ | ✅ |
| Synthetic `HeadExe` (with manifest) | `Exe` | yes | transitive via ProjectReference | `true` ✅ | ✅ |
| Synthetic `ConsumerLib` | `Library` | no | transitive via ProjectReference | `false` ✅ | ✅ |

Inspected `obj/HeadExe.csproj.nuget.g.targets` and confirmed NuGet auto-imports `buildTransitive/Microsoft.Windows.SDK.BuildTools.WinApp.targets` into the consumer head app even though it never directly references the package — proves the MAUI scenario works.

`samples/dotnet-app` (`OutputType=Exe`) was the specific reason the gate uses `OutputType != Library` rather than `OutputType == WinExe` — gating on `WinExe` would have silently broken our own packaged-console-app sample.

## Risks / behavior changes

1. **`DevelopmentDependency` removal** is a behavior change for direct consumers using default `PrivateAssets`. They will now see the package flow transitively unless they set `PrivateAssets="all"` themselves. README updated to recommend `PrivateAssets="all"` for direct head-app consumption.
2. Both `build/` and `buildTransitive/` contain the same files; for direct consumers NuGet imports `build/` (with `buildTransitive/` as fallback), so no double-import. This matches WindowsAppSDK's pattern at scale.

## Not covered by automated tests yet

- Real MAUI multi-target project with iOS/Android TFMs (T3 in the test matrix). The gating logic is based purely on `_WinAppEffectiveTargetPlatformIdentifier != 'Windows'`, which is straightforward and tested at the property-evaluation level, but a workload-installed MAUI build would be the strongest signal. Recommend coordinating with Matthew for a real-world try.
- Release-strategy decisions (version bump, changelog wording, single vs. two-phase release) — leaving for maintainers.


chiaramooney (approved):

src/winapp-NuGet/build/Microsoft.Windows.SDK.BuildTools.WinApp.targets
```diff
@@ -56,12 +64,70 @@
...
+    <_WinAppRunSupportActive Condition="
+        '$(EnableWinAppRunSupport)' == 'true'
+        And '$(WindowsPackageType)' != 'None'
+        And '$(OutputType)' != 'Library'
+        And '$(_WinAppEffectiveTargetPlatformIdentifier)' == 'windows'
+        And (
+          Exists('$(MSBuildProjectDirectory)\Package.appxmanifest')
+          Or Exists('$(MSBuildProjectDirectory)\AppxManifest.xml')
+          Or Exists('$(MSBuildProjectDirectory)\appxmanifest.xml')
+          Or ('$(_WinAppManifestPathUserSet)' == 'true' And Exists('$(WinAppManifestPath)'))
```
mattleibow: In https://github.com/microsoft/winappCli/pull/513/changes#diff-b0415846b8cb2acc21a22ff180ca6cf3c704abd31b07711731d8eb42b3ad9770R52-R65

there are 6 conditional sets of WinAppManifestPath to fist check `$(OutputPath)` then `$(MSBuildProjectDirectory)`. However, in the condition in `_WinAppRunSupportActive` there is just the latter 3.

In a MAUI app, we generate the manifest based on the platform and various msbuild props, so it only exists in the output. This condition is skipped here. If I just check `Exists('$(WinAppManifestPath)')` then it works. But I am not sure what else I break.

nmetulev:
Good catch — fixed in 3d4a637.

You were right that just checking `Exists($(WinAppManifestPath))` works, and the long Or-chain in the gate was actually the source of the asymmetry. After the auto-detection block runs, `$(WinAppManifestPath)` already points at the first existing file across all six supported locations (3 in `$(OutputPath)` + 3 in `$(MSBuildProjectDirectory)`), or at whatever path the consumer set explicitly. So the gate now just reads:

```xml
And '$(WinAppManifestPath)' != ''
And Exists('$(WinAppManifestPath)')
```

The `_WinAppManifestPathUserSet` capture variable that supported the old project-dir-only restriction is gone with it.

The original concern with accepting output-dir paths was "a stale `bin/AppxManifest.xml` from a previous build could re-activate the gate on rebuild," but the other discriminators (`WindowsPackageType != None`, `OutputType != Library`, TPI == windows) already prevent that in projects that aren't intended to be packaged Windows apps — and in a project that is still a packaged Windows app, the gate re-activating is the correct behavior.

I added 3 Pester cases covering the output-dir branch (`Package.appxmanifest`, `AppxManifest.xml`, `appxmanifest.xml`) so MAUI-style head apps stay covered. All 21 tests pass locally. Mind giving the MAUI scenario another try with this build?

PR #522, opened 2026-05-07
ui screenshot: switch to Windows.Graphics.Capture; add --focus flag

nmetulev:
Default `winapp ui screenshot` capture path now uses **Windows.Graphics.Capture (WGC)**, reading the actual DWM-composited window surface. This:

- Preserves rounded corners with true alpha on Win11.
- Removes the black borders previously caused by `GetWindowRect` including the DWM shadow / resize padding.
- Works while the window is occluded by other windows.

PrintWindow is retained as a fallback when WGC is unsupported or fails. `--capture-screen` still does screen-DC BitBlt for popup overlays not owned by the target.

Current behavior:
screenshot

New behavior:
screenshot-new

### New flag: `--focus`
Brings the target window to the foreground before capture, useful when you want the screenshot to match what the user is currently looking at without switching to `--capture-screen` semantics. `--capture-screen` now implies `--focus` and the foregrounding step happens exactly once.

### Implementation notes
- New `WgcCapture` helper using a free-threaded `Direct3D11CaptureFramePool`, B8G8R8A8 surface, copy-to-staging + Map for AOT-safe BGRA32 readback.
- COM interop via `[GeneratedComInterface]` / `[LibraryImport]` to comply with `DisableRuntimeMarshalling` and full trimming.
- Element-crop origin uses `DwmGetWindowAttribute(DWMWA_EXTENDED_FRAME_BOUNDS)` so element bounding rectangles align with the WGC frame's visible-only size.
- Cancellation (Ctrl+C) propagates through the WGC catch instead of falling back to PrintWindow.
- `WgcCapture` is self-contained (its own `IsBlankCapture`); no internal coupling back into `UiAutomationService`.
- PrintWindow blank-frame retry extracted to `CaptureFromWindowWithBlankRetry` shared by the WGC fallback and the no-WGC path.
- COM ownership reviewed: IntPtrs zeroed after `FromAbi` / `ConvertToManaged` so `finally` only releases on error paths.
- Added D3D11 / DXGI / DWM bindings to `NativeMethods.txt`; rooted WGC ABI types in `TrimRoots.xml`.

### CI
`scripts/test-e2e-winui-ui.ps1` now exercises `screenshot --focus` and asserts the PNG. The existing `e2e-test-ui` job on `windows-latest` covers the new default WGC path end-to-end.

### Docs / generated
- Updated `docs/ui-automation.md`, `docs/fragments/skills/winapp-cli/ui-automation.md`, `.github/plugin/agents/winapp.agent.md` to mention `--focus` and describe the new precedence.
- Regenerated `docs/cli-schema.json`, the `ui-automation` SKILL.md, and `src/winapp-npm/src/winapp-commands.ts`.


nmetulev:
Addressed Copilot review feedback in a8a9689:

- **WgcCapture.cs**: Wrap `ID3D11Device`/`ID3D11DeviceContext` in a try/finally so they (and the staging + captured textures inside `CopyFrame`) are disposed after each capture — fixes the COM/GPU leak called out in the device-creation and CopyFrame comments.
- **UiAutomationService.Screenshot.cs**: Replace `Thread.Sleep` in the async foregrounding step with `await Task.Delay(.., ct)` so Ctrl+C / timeouts can interrupt promptly.
- **WgcCapture.cs**: Replace `SDKVersion: 7` with a named `D3D11_SDK_VERSION` constant (CsWin32 doesn't project it).

**Not addressed**: `RunOptions.appArgs` in `winapp-commands.ts` — that file is auto-generated from the CLI schema and the underlying `run` command's hidden `string[]` positional has been wrapped this way for prior commands too; it's a pre-existing generator bug surfaced by regen, not introduced by this PR. Worth tracking in a separate issue against `scripts/generate-commands.mjs`.

zateutsch (approved):

PR #541, opened 2026-05-19
Fix error where update notice checked prerelease versions incorrectly and improperly references cached test output

### LINKED ISSUE #538: [Bug]: Update check shows wrong version in debug builds
### Describe the bug

Image

### To Reproduce

Install a pre-release build and let it check for update

### Expected behavior

If it's pre-release build, it should already be newer than the releases, it should not show an update

### Screenshots

_No response_

### OS Version and details

_No response_

zateutsch:
fixes #538 

Changes:
* Added function to check if a version is prerelease, and to only consider core version when deciding whether or not show update check
* Added check to ensure cached versioning info hasn't been malformed by tests
* Updated corresponding tests

### DIFF
--- a/src/winapp-CLI/WinApp.Cli.Tests/UpdateNotificationGatingTests.cs
+++ b/src/winapp-CLI/WinApp.Cli.Tests/UpdateNotificationGatingTests.cs
@@ -23,6 +23,15 @@ public class UpdateNotificationGatingTests
     private string? _savedCaller;
     private string? _savedUpdateCheck;
 
+    // All environment variable names checked by CIEnvironmentDetectorForTelemetry
+    private static readonly string[] CiVarNames =
+    [
+        "CI", "GITHUB_ACTIONS", "TF_BUILD", "APPVEYOR", "TRAVIS", "CIRCLECI",
+        "TEAMCITY_VERSION", "JB_SPACE_API_URL",
+        "CODEBUILD_BUILD_ID", "AWS_REGION", "BUILD_ID", "BUILD_URL", "PROJECT_ID"
+    ];
+    private Dictionary<string, string?> _savedCiVars = [];
+
     [TestInitialize]
     public void Setup()
     {
@@ -38,15 +47,17 @@ public class UpdateNotificationGatingTests
         _savedCacheDir = Environment.GetEnvironmentVariable("WINAPP_CLI_CACHE_DIRECTORY");
         _savedCaller = Environment.GetEnvironmentVariable("WINAPP_CLI_CALLER");
         _savedUpdateCheck = Environment.GetEnvironmentVariable("WINAPP_CLI_UPDATE_CHECK");
+        _savedCiVars = CiVarNames.ToDictionary(name => name, name => Environment.GetEnvironmentVariable(name));
 
         Environment.SetEnvironmentVariable("WINAPP_CLI_CACHE_DIRECTORY", _tempCacheDir);
         Environment.SetEnvironmentVariable("WINAPP_CLI_CALLER", null);
         Environment.SetEnvironmentVariable("WINAPP_CLI_UPDATE_CHECK", null);
 
         // Clear CI vars to avoid suppression
-        Environment.SetEnvironmentVariable("CI", null);
-        Environment.SetEnvironmentVariable("GITHUB_ACTIONS", null);
-        Environment.SetEnvironmentVariable("TF_BUILD", null);
+        foreach (var name in CiVarNames)
+        {
+            Environment.SetEnvironmentVariable(name, null);
+        }
     }
 
     [TestCleanup]
@@ -55,6 +66,10 @@ public class UpdateNotificationGatingTests
         Environment.SetEnvironmentVariable("WINAPP_CLI_CACHE_DIRECTORY", _savedCacheDir);
         Environment.SetEnvironmentVariable("WINAPP_CLI_CALLER", _savedCaller);
         Environment.SetEnvironmentVariable("WINAPP_CLI_UPDATE_CHECK", _savedUpdateCheck);
+        foreach (var (name, value) in _savedCiVars)
+        {
+            Environment.SetEnvironmentVariable(name, value);
+        }
 
         try { Directory.Delete(_tempCacheDir, recursive: true); } catch { /* best effort */ }
     }
--- a/src/winapp-CLI/WinApp.Cli.Tests/UpdateNotificationGatingTests.cs
+++ b/src/winapp-CLI/WinApp.Cli.Tests/UpdateNotificationGatingTests.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System.Globalization;
+using WinApp.Cli.Services;
 
 namespace WinApp.Cli.Tests;
 
@@ -28,7 +29,7 @@ public class UpdateNotificationGatingTests
         // Create temp cache directory and seed with a "newer" version
         _tempCacheDir = Path.Combine(Path.GetTempPath(), $"winapp_gating_test_{Guid.NewGuid():N}");
         Directory.CreateDirectory(_tempCacheDir);
-        SeedUpdateCheckCache("5.0.0");
+        SeedUpdateCheckCache(GetGuaranteedNewerVersion());
 
         // Create first-run marker so FirstRunService doesn't trigger logging
         File.Create(Path.Combine(_tempCacheDir, ".first-run-complete")).Dispose();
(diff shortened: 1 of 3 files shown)


zateutsch:
@copilot apply changes based on the comments in [this thread](https://github.com/microsoft/winappCli/pull/541#pullrequestreview-4322016892)

nmetulev (commented):
## Copilot PR review — multi-dimensional

**Summary:** Critical 0 · High 1 · Medium 5 · Low 3

| Dimension | Status |
|---|---|
| security | ✓ clean |
| correctness | ⚠ 1 high |
| cli-ux | ⚠ 1 low |
| alternative-solution | ⚠ 2 medium |
| test-coverage | ⚠ 4 findings |
| docs-and-samples | ✓ clean |
| packaging | ✓ clean |
| multi-model (GPT-5.4 cross-check) | 1 confirmed, 1 downgraded |

### Findings at a glance

| ID | File:lines | Domain | Summary |
|---|---|---|---|
| **H1** | `Services/UpdateNotificationService.cs:55-62` | correctness | Cache wipe preserves `LastCheck` → user gets no update notice for ~24h after a poisoned-cache encounter (regresses the fix) |
| M1 | `Tests/UpdateNotificationGatingTests.cs:47-60` | test-coverage | Setup nulls `CI`/`GITHUB_ACTIONS`/`TF_BUILD` without saving them; leaks across in-process tests |
| M2 | `Tests/UpdateNotificationServiceTests.cs:478-499` | test-coverage | `IsUnreasonableVersion` has no test for unparseable (corrupt-cache) input |
| M3 | `Tests/UpdateNotificationServiceTests.cs:586-602` | test-coverage | Wipe test doesn't assert `LastCheck` / `LastShownDate` outcome |
| M4 | `Services/UpdateNotificationService.cs:202-263` | alternative-solution | `GetCoreVersion`/`TryParseMajorVersion` duplicate splitter logic already in `TryParseSemVer` |
| M5 | `Services/UpdateNotificationService.cs:55-67` | alternative-solution | `CheckAndNotify` parses each version string 2-3× across consecutive helpers |
| L1 | `Services/UpdateNotificationService.cs:186-196` | test-coverage / alternative-solution | `IsPreReleaseVersion` is added with 6 unit tests but has **zero** production callers |
| L2 | `Services/UpdateNotificationService.cs:55-62` | cli-ux | Silent cache mutation has no `LogDebug` breadcrumb |
| L3 | `Tests/UpdateNotificationServiceTests.cs` | test-coverage | No test pinning `CurrentVersionProvider` throw → swallow behavior |

---

### Details

#### H1 — `src/winapp-CLI/WinApp.Cli/Services/UpdateNotificationService.cs:55-62` · correctness · **multi-model: confirmed**

When `IsUnreasonableVersion` fires on a poisoned cache (e.g. leftover `999.0.0` from a prior test run), `CheckAndNotify` wipes `LatestVersion` and `LastShownDate` but **preserves `LastCheck`**. The very next gate at lines 77-80 (`!cache.LastCheck.HasValue || (UtcNow - LastCheck).TotalHours >= 24`) then short-circuits to `false`, so the background refresh is skipped.

Net effect: the false notification goes away (good), but the user receives **no real update notification for up to 24 hours** — silently regressing the fix this PR exists to deliver.

```csharp
// Line 60 — LastCheck intentionally omitted:
cache = cache with { LatestVersion = "", LastShownDate = "" };
```

The new test `CheckAndNotify_UnreasonableCachedVersion_DiscardsAndNoNotification` only asserts no notification on the first call and can't catch the missed refresh because `SkipBackgroundRefreshForTesting = true`.

**Recommendation:** Also reset `LastCheck`:
```csharp
cache = cache with { LastCheck = null, LatestVersion = "", LastShownDate = "" };
```
Then extend the test to assert `LastCheck` is null afterward, and ideally add a second test with `SkipBackgroundRefreshForTesting = false` + a stubbed `GetLatestVersionAsync` that verifies the refresh actually runs.

#### M1 — `src/winapp-CLI/WinApp.Cli.Tests/UpdateNotificationGatingTests.cs:47-60` · test-coverage

Setup explicitly nulls `CI` / `GITHUB_ACTIONS` / `TF_BUILD` to defeat `CIEnvironmentDetectorForTelemetry`, but Cleanup only restores `WINAPP_CLI_CACHE_DIRECTORY` / `WINAPP_CLI_CALLER` / `WINAPP_CLI_UPDATE_CHECK`. Contrast `UpdateNotificationServiceTests.cs:22-28,67-70` which snapshots and restores all 13 `CiVarNames`. On CI runners (`CI=true`), this test class permanently clears those vars for the rest of the test process, silently changing the behavior of subsequent CI-detection tests in the same run.

**Recommendation:** Mirror the `_savedCiVars` pattern (or extract it into `BaseCommandTests` / a shared fixture) so both classes snapshot and restore the same set.

#### M2 — `src/winapp-CLI/WinApp.Cli.Tests/UpdateNotificationServiceTests.cs:478-499` · test-coverage

`IsUnreasonableVersion` returns `false` when either side fails to parse. The new tests cover normal / high / test-artifact / boundary / above-threshold / v-prefix, but never exercise the realistic "corrupt cache file" branch (`"garbage"`, `""`, `"1"`, `"1.x"`). That's exactly the failure mode the guard is supposed to backstop.

**Recommendation:** Add `IsUnreasonableVersion_UnparseableCached_ReturnsFalse` and `IsUnreasonableVersion_UnparseableCurrent_ReturnsFalse`, plus a `CheckAndNotify_GarbageCachedVersion_DoesNotNotifyOrThrow` integration test.

#### M3 — `src/winapp-CLI/WinApp.Cli.Tests/UpdateNotificationServiceTests.cs:586-602` · test-coverage

The new wipe test asserts `cache.LatestVersion == ""` but does not assert `LastShownDate` is cleared, and (per H1) does not pin `LastCheck` behavior. Whichever way H1 is resolved, this test should pin the intended `LastCheck` outcome so a future refactor can't silently regress it.

**Recommendation:** Also assert `cache.LastShownDate == ""` and the intended `LastCheck` value (null after H1 fix, or preserved if the team prefers to throttle refreshes).

#### M4 — `src/winapp-CLI/WinApp.Cli/Services/UpdateNotificationService.cs:202-263` · alternative-solution

`GetCoreVersion` (202-212) and `TryParseMajorVersion` (229-240) reimplement the `+`/`-` stripping already inside the nested `TryParseSemVer` local in `IsNewerVersion` (244-263). Three copies of the same SemVer-decomposition rules in one file; future fixes have to be remembered in all three places.

**Recommendation:** Promote `TryParseSemVer` from a nested local to a `private static` method and implement `GetCoreVersion` / `TryParseMajorVersion` in terms of it:
```csharp
return TryParseSemVer(v, out var core, out _) ? core.Major : 0;
```

#### M5 — `src/winapp-CLI/WinApp.Cli/Services/UpdateNotificationService.cs:55-67` · alternative-solution

`CheckAndNotify` calls `IsUnreasonableVersion(latest, current)` then `IsNewerVersion(latest, current)` back-to-back, parsing each string 2-3 times. Once `TryParseSemVer` is promoted (M4), parse each string once and pass `Version` objects to a tiny `IsUnreasonable(Version, Version)` predicate — or inline it, since it has exactly one caller.

#### L1 — `src/winapp-CLI/WinApp.Cli/Services/UpdateNotificationService.cs:186-196` · test-coverage / alternative-solution · **multi-model: downgraded from high → low**

`internal static bool IsPreReleaseVersion(string)` is added with six dedicated unit tests but has **zero production callers**. The PR title says "prerelease versions checked incorrectly", but the actual prerelease behavior is delivered entirely by the unchanged `IsNewerVersion`/`ComparePreRelease` path. A future reader will reasonably assume this helper IS the prerelease fix — it isn't.

**Recommendation:** Delete the method and its six tests. If a future caller actually needs the distinction (e.g. to suppress notifications when on a prerelease against an equal-core stable), add it back together with the production code path that uses it.

#### L2 — `src/winapp-CLI/WinApp.Cli/Services/UpdateNotificationService.cs:55-62` · cli-ux

The cache-wipe branch silently mutates user state on disk with no log entry. A user troubleshooting under `--verbose` has no breadcrumb. The service already has `ILogger<UpdateNotificationService>` injected, so the cost is one line:
```csharp
logger.LogDebug("Discarded unreasonable cached update version {Version}", cache.LatestVersion);
```

#### L3 — `src/winapp-CLI/WinApp.Cli.Tests/UpdateNotificationServiceTests.cs` · test-coverage

The new `CurrentVersionProvider` test seam is invoked inside `CheckAndNotify`'s silent outer `try`/`catch`. No test pins "provider throws → no crash, no cache mutation, no notification".

**Recommendation:** Add `CheckAndNotify_CurrentVersionProviderThrows_DoesNotCrashAndDoesNotNotify`.

---

### Coverage notes

- **security:** Verified no new `Process.Start` / HTTP / cert / manifest / regex / deserialization / `NoWarn` surface. Cache path and atomic-write pattern unchanged. New helpers operate on local strings only.
- **docs-and-samples:** No `Commands/*.cs` changed → `cli-schema.json` regen not needed. `docs/usage.md` "Update Checks" section documents the opt-out env var only; no described behavior contradicted by this PR.
- **packaging:** Lands uniformly in all distribution channels (native / npm / NuGet / VSC) via the shared CLI binary. No version bump needed (Nerdbank.GitVersioning is height-based).
- **multi-model:** GPT-5.4 cross-checked H1 (confirmed) and the original H2 (downgraded to L1). Independently scanned the static `_refreshScheduled` interlock — no additional critical/high findings.

---

**Blocker:** **H1** — without also resetting `LastCheck`, the fix suppresses the false notification but then suppresses real ones too for ~24h.

nmetulev (approved):

PR #550, opened 2026-05-27
Fix RCE via execution alias resolved against CWD/PATH

nmetulev:
## Summary

Fixes an MSRC-reported RCE (case `bbe50a79-0358-f111-8ce8-000d3ac30a19`) in `winapp run --with-alias`.

`winapp run --with-alias` passed the alias from `<uap5:ExecutionAlias>` verbatim to `ProcessStartInfo.FileName` with `UseShellExecute = false`. .NET dispatches that through `CreateProcessW`, which resolves bare filenames against the **current working directory before PATH**. The Windows App Execution Alias proxy lives under `%LOCALAPPDATA%\Microsoft\WindowsApps\` (on PATH), but a hostile `a.exe` shipped alongside `Package.appxmanifest` in an untrusted repo would win the lookup and run in place of the registered alias — RCE with the developer's privileges from a clone-and-run flow.

## Fix

Add `Helpers/ExecutionAliasResolver.cs` with:

- `IsSafeAliasName(alias)` — validates the alias is a bare filename: rejects path separators (`\`, `/`), `..`, rooted paths (drive letters, UNC, leading separator), invalid filename chars (incl. NUL/control chars), oversized names, dot-names.
- `ResolveAliasPath(alias, baseDirectory?)` — combines the validated alias with the WindowsApps base directory and returns the absolute `FileInfo`. `baseDirectory` is parameterised for testability; production callers use `%LOCALAPPDATA%\Microsoft\WindowsApps`.

`RunCommand.LaunchViaExecutionAliasAsync` now:

1. Validates the extracted alias via `IsSafeAliasName`; rejects path-like values with a clear log message.
2. Resolves it to the absolute `%LOCALAPPDATA%\Microsoft\WindowsApps\<alias>` path.
3. Verifies the proxy file exists; if not, errors with a message pointing the user at AUMID launch (drop `--with-alias`).
4. Passes that **absolute path** to `ProcessStartInfo.FileName`. Absolute paths bypass `CreateProcess`'s CWD/PATH search entirely, so a hostile binary next to the manifest can no longer hijack the launch.

`MsixService.ExtractExecutionAliases` and `BuildAliasProcessStartInfo` keep their original semantics — validation/resolution is a launch concern, not a manifest-parsing one.

## Tests

New `ExecutionAliasResolverTests.cs` (31 tests):

- ✅ Accepted bare names: simple `.exe`, mixed case, hyphen/underscore/digits, single letter, no extension, internal dots
- ❌ Rejected: null/empty/whitespace/tab, `.`, `..`, all path separator variants, parent traversal (both slash flavours), nested paths, leading `/` and `\`, absolute paths with drive letters, drive-relative, UNC, extended-length UNC, NUL char, invalid filename chars (`* ? < > | " :`), names > 255 chars
- ✅ `ResolveAliasPath`: returns correct path under custom + default WindowsApps base; returns null for hostile aliases; `FileInfo.Exists` correctly reflects filesystem state for present/missing proxies

## Verification

- `dotnet build src/winapp-CLI/winapp.sln -c Debug` — clean (0 warnings, 0 errors)
- 1118 / 1118 unit tests pass (5 unrelated E2E failures need `npm run build` for the Node harness — not exercised by this change)
- `.\scripts\build-cli.ps1 -SkipMsix -SkipNpm -SkipTests` — full pipeline succeeds; auto-generated `docs/cli-schema.json` and skill docs unchanged (no user-facing option text changed)

## Notes

- The user-facing `--with-alias` option description is unchanged; this is an internal correctness/security fix, not an API change.
- `BuildAliasProcessStartInfo` is intentionally untouched — it just plumbs a string into `ProcessStartInfo`; the call site now passes a fully-qualified path.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/samples/cpp-app/test.Tests.ps1
+++ b/samples/cpp-app/test.Tests.ps1
@@ -100,6 +100,10 @@ add_executable(test-cpp-app main.cpp)
             Invoke-WinappCommand -Arguments "run build\Debug --unregister-on-exit"
         }
 
+        It "runs app via execution alias with winapp run --with-alias" -Skip:$script:skip {
+            Invoke-WinappCommand -Arguments "run build\Debug --with-alias --unregister-on-exit"
+        }
+
         It "CMake builds release successfully" -Skip:$script:skip {
             $output = cmake --build build --config Release 2>&1
             $LASTEXITCODE | Should -Be 0 -Because "CMake build failed: $output"
--- a/samples/rust-app/test.Tests.ps1
+++ b/samples/rust-app/test.Tests.ps1
@@ -83,6 +83,27 @@ Describe "Rust App Sample" {
             } finally { Pop-Location }
         }
 
+        It "Should reject hostile execution alias in manifest" -Skip:$script:skip {
+            # Regression test for MSRC RCE fix: a manifest containing a non-bare /
+            # path-traversal ExecutionAlias must be rejected by `winapp run --with-alias`
+            # before any process is launched.
+            Push-Location $script:rustProjectDir
+            try {
+                $manifestPath = Join-Path $script:rustProjectDir "Package.appxmanifest"
+                $original = Get-Content -Raw -Path $manifestPath
+                try {
+                    $hostile = $original -replace 'Alias="[^"]+"', 'Alias="..\evil.exe"'
+                    Set-Content -Path $manifestPath -Value $hostile -NoNewline
+
+                    $output = & winapp run .\target\debug --with-alias --unregister-on-exit 2>&1
+                    $LASTEXITCODE | Should -Not -Be 0 -Because "hostile alias must be rejected"
+                    ($output -join "`n") | Should -Match '(?i)alias' -Because "error should mention alias validation. Got: $output"
+                } finally {
+                    Set-Content -Path $manifestPath -Value $original -NoNewline
+                }
+            } finally { Pop-Location }
+        }
+
         It "Should build Rust app in release mode" -Skip:$script:skip {
             Push-Location $script:rustProjectDir
             try {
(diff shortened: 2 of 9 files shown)


zateutsch (commented):
PR review mostly clean - this is the only thing that popped up that might be worth double-checking:
 
 ## M1  src/winapp-CLI/WinApp.Cli/Services/ManifestService.cs:561-570
 - Severity: medium
 - Confidence: high
 - Domain: correctness
 - Multi-model: confirmed
 - Finding: `manifest add-alias` now rejects manifests with `Executable="app\my-app.exe"` because the raw Executable value is validated as a bare filename before extracting the leaf name.
 - Evidence: ManifestService infers aliasName from Application/@Executable (e.g. `app\my-app.exe`) then passes it to `IsSafeAliasName()` which rejects any string containing path separators. This is a documented manifest shape (Electron guide).
 - Recommendation: When `--name` is omitted, apply `Path.GetFileName(executable)` to extract the leaf before validation. Still reject truly hostile patterns like `..\evil.exe` (GetFileName would return `evil.exe` — so also reject if the original path contains `..` segments).

nmetulev:
Good catch — fixed in ceff2e6b.

Followed your recommendation: when `--name` is omitted, `Path.GetFileName(executable)` is applied before validation so path-prefixed Executable values like `app\my-app.exe` are accepted (alias becomes `my-app.exe`). To prevent `GetFileName` from silently stripping path traversal out of a hostile manifest, the inferred path also fails fast with `InvalidAliasName` if any segment is literally `..`.

Added `AddAlias_PathPrefixedExecutable_InfersLeafFilenameAsAlias` with DataRows for `app\my-app.exe`, `app/my-app.exe`, and `bin\sub\nested.exe` to lock in the Electron pattern. The existing `AddAlias_UnsafeExecutableInferred_RejectsAndDoesNotModifyManifest` test (Executable=`..\evil.exe`) still passes via the new `..` segment guard. 33/33 ManifestAddAlias tests pass.

zateutsch (approved):

PR #558, opened 2026-05-31
Fix update check timeout bug

### LINKED ISSUE #555: Update-check cache (`~/.winapp/.update-check`) not reliably written on first run — 250 ms timeout races GitHub API
## Summary

`~/.winapp/.update-check` is not reliably written on first run (or after the cache file is deleted). The background refresh task is fire-and-forget with only a **250 ms** foreground wait, so when the GitHub API round-trip exceeds that, the process exits before `RefreshCacheAsync` completes its atomic `File.Move`, and no cache is persisted.

This means:
- First-time users may not see the `.update-check` file appear at all after their first command.
- The 24-hour gate that depends on `LastCheck` never trips on subsequent runs (because there is no `LastCheck`), so we re-attempt — and re-race — on every invocation until one finally wins.

## Environment

- winapp CLI version: `0.3.2-prerelease.23`
- OS: Windows 11
- Install path: `C:\Users\<user>\AppData\Local\Microsoft\WindowsApps\winapp.exe` (MSIX wrapper)
- Network: warm connection, no VPN
- No `WINAPP_CLI_UPDATE_CHECK` opt-out, no CI env vars set

## Repro steps

```powershell
$f = "$env:USERPROFILE\.winapp\.update-check"
$results = @()
for ($i = 1; $i -le 10; $i++) {
    if (Test-Path $f) { Remove-Item -Force $f }
    $sw = [Diagnostics.Stopwatch]::StartNew()
    winapp --version 2>$null | Out-Null
    $sw.Stop()
    $immediate = Test-Path $f
    Start-Sleep -Milliseconds 1500   # generous grace period for any straggling background write
    $after = Test-Path $f
    $results += [pscustomobject]@{
        Iter           = $i
        CmdMs          = [int]$sw.ElapsedMilliseconds
        CacheOnExit    = $immediate
        CacheAfter1_5s = $after
    }
}
$results | Format-Table -AutoSize
"Cache missing after 1.5s grace: $(($results | Where-Object { -not $_.CacheAfter1_5s }).Count) / 10"
```

## Actual results

```
Iter CmdMs CacheOnExit CacheAfter1_5s
---- ----- ----------- --------------
   1   651       False          False
   2   332        True           True
   3   430       False          False
   4   684       False          False
   5   387        True           True
   6   657        True           True
   7   608       False          False
   8   405        True           True
   9   375        True           True
  10   678        True           True

Cache missing after 1.5s grace: 4 / 10
```

**40% of invocations leave no cache file**, even after a 1.5 s grace period post-exit. The correlation between `CmdMs` and `CacheAfter1_5s` is the smoking gun:

- Fast runs (~330–430 ms) → cache written ✓
- Slow runs (~600–680 ms) → cache missing ✗

When GitHub's response exceeds the **250 ms** `FirstRunRefreshTimeout`, the foreground wait expires, the process tears down, and the in-flight HTTP / `File.Move` is killed.

## Expected results

After any successful invocation with network access and no opt-out, `~/.winapp/.update-check` should exist and contain a valid `LastCheck` timestamp, so subsequent invocations honor the 24-hour gate.

## Root cause

`src/winapp-CLI/WinApp.Cli/Services/UpdateNotificationService.cs`

```csharp
private static readonly TimeSpan FirstRunRefreshTimeout = TimeSpan.FromMilliseconds(250); // line 25

// ... line 78–104:
var refreshTask = Task.Run(async () =>
{
    try { await RefreshCacheAsync(cacheFile); }
    catch { /* swallowed */ }
    finally { Interlocked.Exchange(ref _refreshScheduled, NotScheduled); }
});

if (!cache.LastCheck.HasValue)
{
    refreshTask.Wait(FirstRunRefreshTimeout); // only 250 ms!
}
```

If `RefreshCacheAsync` (HTTPS GET to `api.github.com` + JSON parse + `File.WriteAllText` + `File.Move`) doesn't complete in 250 ms, the process exits while the task is still in flight. There is no flush, no shutdown await, no retry — the failure is also silently swallowed.

## Suggested fix (any one of these)

1. **Bump the timeout** to ~1500 ms — long enough for a warm GitHub round-trip on most connections, while still imperceptible if the user hits a hot path.
2. **Persist a placeholder synchronously** before the HTTP call:
   ```csharp
   WriteCacheFile(cacheFile, new UpdateCheckCache(DateTimeOffset.UtcNow, "", existingCache.LastShownDate));
   ```
   This makes the 24-hour gate trip on the next run even if HTTP loses the race, eliminating the "race every invocation" amplifier.
3. **Make `CheckAndNotify` async** and `await` the refresh task during host shutdown so it gets a real chance to complete.

(2) is the lowest-risk, smallest-diff fix and also fixes the "every invocation re-races" amplifier; (1) is complementary.

## Impact

- New users may be confused that the file documented to exist doesn't appear.
- Users on slower networks / cold DNS / behind proxies will hit this far more often than the 40% observed locally on a warm connection.
- The 24-hour throttle is effectively defeated until a fast round-trip happens to win, increasing GitHub API call volume.

## Workarounds

- Run `winapp` a few times; eventually a fast response wins and the cache is created.
- Or set `WINAPP_CLI_UPDATE_CHECK=0` to skip the check entirely (also skips notifications).

zateutsch:
fixes #555 

Upped timeout on github request from 250ms to 1s
Writing a placeholder value to update check file to ensure there aren't subsequent failures

### DIFF
--- a/.github/workflows/test-samples.yml
+++ b/.github/workflows/test-samples.yml
@@ -37,7 +37,7 @@ jobs:
         dotnet-version: '10.0.x'
     - uses: actions/setup-node@v5
       with:
-        node-version: '24'
+        node-version: '22'
     - name: Build npm package
       shell: pwsh
       run: .\scripts\build-cli.ps1 -SkipTests -SkipMsix -SkipVsc
@@ -127,7 +127,7 @@ jobs:
         contains(fromJson('["electron", "tauri-app", "cpp-app", "dotnet-app", "wpf-app", "rust-app", "flutter-app", "packaging-cli"]'), matrix.sample)
       uses: actions/setup-node@v5
       with:
-        node-version: '24'
+        node-version: '22'
 
     - name: Setup Flutter
       if: >-
--- a/src/winapp-CLI/WinApp.Cli/Services/UpdateNotificationService.cs
+++ b/src/winapp-CLI/WinApp.Cli/Services/UpdateNotificationService.cs
@@ -22,7 +22,7 @@ internal class UpdateNotificationService(
     private const string UpdateCheckFileName = ".update-check";
     private const int CheckIntervalHours = 24;
     private const int MaxReasonableFutureMajorDelta = 20;
-    private static readonly TimeSpan FirstRunRefreshTimeout = TimeSpan.FromMilliseconds(250);
+    private static readonly TimeSpan FirstRunRefreshTimeout = TimeSpan.FromMilliseconds(1000);
 
     // For testing only — when true, skips the fire-and-forget background network refresh
     internal bool SkipBackgroundRefreshForTesting;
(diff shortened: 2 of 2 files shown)


Jaylyn-Barbee (approved):
LGTM.

PR #568, opened 2026-06-03
Fix: Wingetcreate installer count mismatch in release pipeline

zateutsch:
Winget release failed because two portable installers were added to our manifest for 0.3.1, which brought the total installer count up to 4.

Our script was only passing two url's, causing winget to error out on an installer count mismatch.


Jaylyn-Barbee (approved):
LGTM

PR #576, opened 2026-06-12
Screenshot command with an out directory that doesn't exist will create that directory instead of throwing an exception

### LINKED ISSUE #572: winapp ui screenshot fails (IO_PathNotFound) when output parent directory doesn't exist — should create parent dirs
### Summary
`winapp ui screenshot -o <path>` fails with `IO_PathNotFound_Path` when the **parent directory of the output path does not exist**. It would be much friendlier to auto-create the parent directory (like `mkdir -p`), or at minimum emit a clear "output directory does not exist" message rather than a raw `IO_PathNotFound_Path`.

### Environment
- winapp version: **0.3.1**
- OS: Windows 11

### Repro
```powershell
# Pick any visible window hwnd
$h = (winapp ui list-windows --json | ConvertFrom-Json | Select-Object -First 1).hwnd

# Output path whose parent folder does NOT exist:
winapp ui screenshot -w $h -o "$env:TEMP\does-not-exist-yet\nested\shot.png"
```

### Actual
```
❌ IO_PathNotFound_Path, C:\Users\<user>\AppData\Local\Temp\does-not-exist-yet\nested\shot.png
```
No file is written.

### Expected
One of:
1. **(preferred)** The command creates missing parent directories and writes the PNG (parity with how most CLI capture tools behave), or
2. A clear, actionable error such as `Output directory 'C:\...\does-not-exist-yet\nested' does not exist`.

### Workaround
Pre-create the directory first:
```powershell
New-Item -ItemType Directory -Force -Path (Split-Path $out) | Out-Null
winapp ui screenshot -w $h -o $out   # works
```

### Why it matters
In batch/automation scripts that lay out artifacts under per-test folders (e.g. `artifacts/L394/step-01.png`), the screenshot silently fails for the first capture in each new folder unless every directory is pre-created. The raw `IO_PathNotFound_Path` text doesn't make the cause obvious.

zateutsch:
fixes #572 

If a user specifies an out path with `-o` that contains directory that doesn't exist, create it.

### DIFF
--- a/src/winapp-CLI/WinApp.Cli/Commands/UiScreenshotCommand.cs
+++ b/src/winapp-CLI/WinApp.Cli/Commands/UiScreenshotCommand.cs
@@ -97,6 +97,11 @@ internal class UiScreenshotCommand : Command, IShortDescription
                 var pngBytes = EncodePng(pixels, w, h);
 
                 var filePath = output ?? "screenshot.png";
+                var dir = Path.GetDirectoryName(Path.GetFullPath(filePath));
+                if (dir is not null)
+                {
+                    Directory.CreateDirectory(dir);
+                }
                 await File.WriteAllBytesAsync(filePath, pngBytes, cancellationToken);
                 var absolutePath = Path.GetFullPath(filePath);
 
@@ -217,6 +222,11 @@ internal class UiScreenshotCommand : Command, IShortDescription
 
             // Compose all captures side-by-side into single image
             var pngBytes = ComposeSideBySide(captures);
+            var dir = Path.GetDirectoryName(Path.GetFullPath(filePath));
+            if (dir is not null)
+            {
+                Directory.CreateDirectory(dir);
+            }
             await File.WriteAllBytesAsync(filePath, pngBytes, ct);
             var absolutePath = Path.GetFullPath(filePath);
 


src/winapp-CLI/WinApp.Cli/Commands/UiScreenshotCommand.cs
```diff
@@ -97,6 +97,11 @@ public override async Task<int> InvokeAsync(ParseResult parseResult, Cancellatio
                 var pngBytes = EncodePng(pixels, w, h);
 
                 var filePath = output ?? "screenshot.png";
+                var dir = Path.GetDirectoryName(Path.GetFullPath(filePath));
+                if (dir is not null)
+                {
+                    Directory.CreateDirectory(dir);
```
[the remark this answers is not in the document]
zateutsch: >  I'd say no — this is a one-liner call to a well-known framework API (Directory.CreateDirectory), and the screenshot command is inherently integration-level (needs a live window + UI Automation). A unit test would just be verifying that .NET's Directory.CreateDirectory works, which isn't valuable.
> If you already had an integration test that exercises the screenshot path end-to-end, it'd be nice to add a subcase with a nested output path, but I wouldn't create new test infrastructure just for this.

src/winapp-CLI/WinApp.Cli/Commands/UiScreenshotCommand.cs
```diff
@@ -217,6 +222,11 @@ private async Task<int> CaptureMultipleWindows(
 
             // Compose all captures side-by-side into single image
             var pngBytes = ComposeSideBySide(captures);
+            var dir = Path.GetDirectoryName(Path.GetFullPath(filePath));
+            if (dir is not null)
+            {
+                Directory.CreateDirectory(dir);
+            }
```
chiaramooney, line 229: Maybe I'm reading the code wrong but do we need to check if the directory already exists before we try creating it? Or does the CreateDirectory function handle that check for us?

zateutsch: CreateDirectory handles that check.

chiaramooney (approved):

PR #596, opened 2026-06-30
Add `--update-csproj` to `winapp manifest add-alias`

### LINKED ISSUE #520: [Feature]: winapp manifest add-alias should have a flag that automatically adds or updates WinAppRunUseExecutionAlias prop

No.

I think it would make documentation/blogs/demos more powerful if we could support fully enabling an execution alias with a one-liner.

`winapp manifest add-alias --update-csproj` for example.

### Additional context

_No response_

zateutsch:
fixes #520 

Adds an opt-in `--update-csproj` flag that sets `<WinAppRunUseExecutionAlias>true</WinAppRunUseExecutionAlias>` in the `.csproj` next to the manifest, so `winapp run` / `dotnet run` launch the app through its execution alias and keep console I/O in the current terminal.

### What's included
- **New flag** `--update-csproj` on `manifest add-alias`. Default off; behavior is fully additive.
- **Comment-aware csproj editing.** Property and anchor (`RuntimeIdentifier` / `TargetFramework` / `<PropertyGroup>`) matches inside XML comments are skipped, so commented-out config is never spliced into the live project. Shared `EnsureBooleanPropertyTrueAsync` helper now backs both `EnsureWinAppRunUseExecutionAliasAsync` and the existing `EnsureEnableMsixToolingAsync`.
- **Accurate reporting.** New `CsprojPropertyOutcome` (`Added` / `Updated` / `AlreadyTrue` / `NotModified`) drives precise logging. The alias is still written, but the command exits non-zero when `--update-csproj` is requested and no `.csproj` / insertion point is found, instead of silently no-op-ing.
- **Docs** updated: `usage.md`, `dotnet-run-support.md`, and the `manifest` skill fragment.

### Testing
- 6 new edge-case tests (commented-out-property regression, already-true no-op, `<TargetFrameworks>`-only, no-`<PropertyGroup>` error, multiple csproj, already-exists alias + flag).
- Full suite: **1334 passed, 1 env-skip, 0 failed**. Release build clean (analyzers pass).

### Notes
- No breaking changes — flag is opt-in and `IDotNetService` is internal.
- Autogenerated docs (`cli-schema.json`, plugin/`.claude` SKILL.md) are regenerated by `scripts/build-cli.ps1` during the official build.

### DIFF
--- a/.claude/skills/winapp-manifest/SKILL.md
+++ b/.claude/skills/winapp-manifest/SKILL.md
@@ -103,7 +103,7 @@ winapp manifest add-alias --update-csproj
 
 This adds a `uap5:AppExecutionAlias` extension to the manifest. If the alias already exists, the command reports it and exits successfully.
 
-With `--update-csproj`, the command also sets `<WinAppRunUseExecutionAlias>true</WinAppRunUseExecutionAlias>` in the `.csproj` next to the manifest (adding or updating the property), so `winapp run` / `dotnet run` launch the app through the alias and keep console I/O in the current terminal.
+With `--update-csproj`, the command also sets `<WinAppRunUseExecutionAlias>true</WinAppRunUseExecutionAlias>` in the `.csproj` next to the manifest (adding or updating the property), so `winapp run` / `dotnet run` launch the app through the alias and keep console I/O in the current terminal. The alias is still added to the manifest, but the command exits non-zero if it cannot find a `.csproj` or insert the property, so the request is never silently dropped.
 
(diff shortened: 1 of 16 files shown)


nmetulev (commented):
Shouldn't it check if the winapp nuget project is also added? Otherwise the property does nothing. But I also wonder if we are trying to do too much here but not sure....

nmetulev:
I'm not convinced we need this feature - there is not a real user pain point here, and I worry the code adds complexity we have to maintain just so it can make demos/blogs easier.

PR #597, opened 2026-07-01
Enhance `winapp run --debug-output` crash stacks with WinUI stowed-exception triage

### LINKED ISSUE #549: Enrich winapp run --debug-output crash stacks for WinUI apps
## What

When `winapp run --debug-output` captures a crash dump from a WinUI app, the resulting stack is fairly thin — typically the user's faulting frame plus a couple of cswinrt ABI shims. There's no visibility into the XAML dispatcher / CoreMessaging / runtime chain that actually led to the failure, and no breakdown of stowed exceptions (`0xC000027B`) which are the most common WinUI crash shape.

## Why

Most WinUI crashes originate inside a XAML event handler and surface as a stowed exception wrapping a managed exception. The minimum context a developer needs to diagnose these is:

- The originating HRESULT and its ErrorContext chain
- The full native dispatch stack (Microsoft.UI.Xaml → CXcpDispatcher → CoreMessagingXP → CLR host)
- The managed user frame that threw

Today only the last one is reliably produced.

## Potential approach

The WinUI team ships a WinDbg JavaScript extension (`winui-dbgext.js` in `microsoft/microsoft-ui-xaml`) that already produces exactly this output via `!xamltriage` and `!xamlstowed`. It runs against a minidump and works well with public symbols.

A viable integration shape that avoids requiring a WinDbg install:

1. Host DbgEng directly via the `Microsoft.Debugging.Platform.DbgEng` NuGet package (winapp already uses `Microsoft.Diagnostics.Runtime.Utilities` for its native fallback, so the API surface is familiar).
2. Add the `Microsoft.Debugging.Platform.SymSrv` package and **co-locate `symsrv.dll` next to `dbgeng.dll`** — without this, `srv*` paths silently no-op and symbols never download. This is the main gotcha.
3. Add `Microsoft.Debugging.Platform.DbgX` for `dbgmodel.dll` / `msdia140.dll`.
4. Acquire `JsProvider.dll` (the JS scripting host for `.scriptload`) — this is **not** distributed on NuGet; it ships only with Debugging Tools for Windows. Options: bundle it, or download a WinDbg/SDK package on first use.
5. Fetch `winui-dbgext.js` from the public WinUI repo (pin a SHA).
6. After the existing ClrMD pass, run `.scriptload winui-dbgext.js; !xamlstowed; !xamltriage` against the same dump and append the output to the debug log.

Gating it behind a flag (or auto-enabling when `Microsoft.UI.Xaml.dll` is in the module list) keeps it scoped to WinUI apps.

## Notes for whoever picks this up

- `!xamlstowed` produces the most valuable output and works fully with the NuGet + JsProvider setup above.
- `!xamltriage`'s managed-frame walker depends on SOS / CLR data access being loaded in the session; without that it logs `CORDBG_E_MISSING_METADATA` for those frames. The headline summary (error code, message, user frames, ErrorContext) still comes through. Easiest path is probably to combine winapp's existing ClrMD managed stack with the extension's native output.
- Relevant existing code: `src/winapp-CLI/WinApp.Cli/Services/CrashDumpService.cs` (`AnalyzeWithDbgEng`) is where the new pass would slot in.

zateutsch:
## Enhance `winapp run --debug-output` crash stacks with WinUI stowed-exception triage

Closes #549

### What

When `winapp run --debug-output` captures a crash from a WinUI 3 app, the resulting stack was thin — the user's faulting frame plus a couple of cswinrt ABI shims, with no visibility into the XAML dispatcher / CoreMessaging / runtime chain that led to the failure, and no breakdown of stowed exceptions (`0xC000027B`), the most common WinUI crash shape.

This PR adds an automatic **WinUI stowed-exception triage pass** on top of the existing ClrMD analysis. When `Microsoft.UI.Xaml.dll` is present in the crashed process's module list, winapp now surfaces:

- The originating HRESULT and its **ErrorContext** chain
- The full native XAML dispatch stack (`Microsoft.UI.Xaml` → `CXcpDispatcher` → `CoreMessagingXP` → CLR host)
- The managed user frame that threw (from winapp's existing ClrMD pass)

The feature is **auto-enabled** for WinUI apps — no new flag. The standard managed/native analysis is unchanged for all other apps.

### How

- Hosts **DbgEng** directly via the `Microsoft.Debugging.Platform.DbgEng` NuGet package and runs the WinUI team's `winui-dbgext.js` extension (`!xamlstowed` / `!xamltriage`) against the same minidump — no WinDbg install required.
- Co-locates `symsrv.dll` (from `Microsoft.Debugging.Platform.SymSrv`) next to `dbgeng.dll` so `srv*` symbol paths resolve.
- `JsProvider.dll` (the JS scripting host, **not** on NuGet) is fetched on first use directly from the official WinDbg download using ranged reads (only the few hundred KB needed), Authenticode-verified as Microsoft-signed, and cached under the winapp global directory.
- The extension script is pinned by SHA and hash-gated before load.
- Debugger component versions are **pinned** (kept in sync with `Directory.Packages.props` via a drift test) so dev/CI acquisition is deterministic.
- Everything is cached, so subsequent runs are offline. For locked-down environments, set `WINAPP_DBGTOOLS_DIR` to a directory containing `dbgeng.dll` and `JsProvider.dll`. When the binaries can't be obtained, triage is skipped with a clear log note and the standard analysis still runs.

`--symbols` additionally improves the native XAML dispatch stack for WinUI apps.

### Testing

- Added unit tests covering: ZIP64 central-directory parsing (ranged extraction), stowed-exception record selection, NuGet global-cache version pinning + fallback, child-process argument construction, the pinned extension-hash gate, and a drift test tying the pinned debugger version to `Directory.Packages.props`.
- Full suite: **1355 passed, 0 failed** (1 environment-only skip). Build clean with 0 warnings, including the Release/AOT (ILC) compile.

### Docs

- Updated `docs/usage.md` (`--debug-output` / `--symbols`) and the `setup.md` skill fragment; `docs/debugging.md` documents the triage pass and `WINAPP_DBGTOOLS_DIR` override. Regenerated Copilot/Claude skills.

### Notes

- `!xamlstowed` produces the highest-value output and works fully with the NuGet + JsProvider setup. `!xamltriage`'s managed-frame walker depends on SOS/CLR data access; winapp instead combines its own ClrMD managed stack with the extension's native output, so managed frames are always resolved.
- Relevant new code: `Services/XamlTriageService.cs`, `XamlTriageBinaries.cs`, `XamlTriageRunner.cs`, `WinDbgJsProviderAcquirer.cs`, `Helpers/ZipRangeExtractor.cs`, `Helpers/AuthenticodeVerifier.cs`; triage slots in after `CrashDumpService.AnalyzeWithClrMD`.


src/winapp-CLI/WinApp.Cli/Services/CrashDumpService.cs
```diff
@@ -142,22 +163,64 @@ internal sealed class CrashDumpService(IAnsiConsole console, ILogger<CrashDumpSe
...
+            // WinUI triage pass — auto-enabled when Microsoft.UI.Xaml.dll is in the dump's
+            // module list. Recovers the stowed exception (0xC000027B) and the XAML dispatch
+            // chain that the ClrMD/DbgEng passes alone cannot surface.
+            string? xamlTriage = null;
+            if (isWinUi)
+            {
+                console.MarkupLine(useSymbols
+                    ? "[dim]Running WinUI stowed-exception triage (first run may download debugger components and symbols; this can take a few minutes)...[/]"
+                    : "[dim]Running WinUI stowed-exception triage (first run may download debugger components)...[/]");
+                xamlTriage = await xamlTriageService.TryAnalyzeAsync(dumpPath, useSymbols);
```
nmetulev: **H1 (High) — triage failure discards the managed crash stack.**

This call is unguarded and runs *before* the managed summary (computed at line 191) is emitted. `XamlTriageService.TryAnalyzeAsync` rethrows **all** `OperationCanceledException` (see `XamlTriageService.cs:101`) — which includes `TaskCanceledException` from `HttpClient.Timeout` (100 s here / 5 min for the ~23 MB `JsProvider.dll` download). That rethrow propagates out of this line into the outer `catch (Exception)` at line 296, which prints *"Analysis failed. Open in WinDbg"* and returns — **throwing away the ClrMD managed stack that was already computed and that worked fine before this PR.**

This is a regression on exactly the first-run / slow / blocked-network scenario the graceful-skip design targets. Verified against source and confirmed by a second-model cross-check. (Not force-reproduced live — it needs an induced network timeout — but the 23 MB first-run download makes it realistic.)

**Fix:** guard this call locally (log + continue with `xamlTriage = null`), and/or in `TryAnalyzeAsync` only rethrow OCE when the caller's token is actually cancelled — no `CancellationToken` is even passed here today, so 100% of OCEs reaching that handler are internal timeouts that should be swallowed.

src/winapp-CLI/WinApp.Cli/Services/XamlTriageService.cs
```diff
@@ -0,0 +1,344 @@
...
+                  $"engine ({binaries.Source}), so symbols could not be downloaded. Install Debugging " +
+                  $"Tools for Windows or point {XamlTriageBinaries.EnvOverride} at a layout that includes symsrv.dll."
+                : null;
+
+            var gapNote = DescribeSymbolGap(output, useSymbols);
+            var notes = string.Join("\n", new[] { symbolNote, gapNote }.Where(n => n != null));
+            return notes.Length == 0
+                ? $"{header}\n{output.Trim()}"
+                : $"{header}\n{notes}\n\n{output.Trim()}";
+        }
```
nmetulev, line 105: **H1 root cause.** `catch (OperationCanceledException) { throw; }` rethrows internal `HttpClient.Timeout` cancellations (`TaskCanceledException`) identically to genuine caller cancellation — but this method is invoked with no `CancellationToken` (`CrashDumpService.cs:202`), so every OCE that reaches here is an internal timeout. The rethrow defeats this method's own fail-open contract (the general `catch` just below returns `null`) and lets the caller's outer handler discard the managed stack.

**Fix:** `catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested) { throw; }` and let internal timeouts fall through to the fail-open `return null;` path.

src/winapp-CLI/WinApp.Cli/Services/XamlTriageBinaries.cs
```diff
@@ -0,0 +1,380 @@
...
+        var id = package.ToLowerInvariant();
+        var version = await ResolveDownloadVersionAsync(http, id, pinnedVersion, logger, cancellationToken);
+        if (string.IsNullOrEmpty(version))
+        {
+            return false;
+        }
+
+        // Download and extract the .nupkg (a zip archive) to a temp directory.
+        var nupkgUrl = $"{FlatContainer}/{id}/{version}/{id}.{version}.nupkg";
+        using var nupkgResponse = await http.GetAsync(nupkgUrl, cancellationToken);
```
nmetulev, line 467: **M4 (Medium) — engine package downloaded without integrity verification.** The `.nupkg` is fetched over HTTPS from the flat-container feed and extracted (line 300) with no package hash / signature check, and its native DLLs are then loaded into the debugger process (`XamlTriageRunner`). HTTPS protects transport but not a compromised or mirrored feed.

**Consider** verifying the downloaded package content hash against an expected value (or otherwise pinning integrity). Note: `NugetService` also skips this, so simply reusing it would not close the gap.

src/winapp-CLI/WinApp.Cli/Services/XamlTriageBinaries.cs
```diff
@@ -0,0 +1,380 @@
...
+            if (resolved != null)
+            {
+                logger.LogDebug("Resolved WinUI triage debugging binaries from {Source}: {Dir}", source, dir);
+                return resolved;
+            }
+        }
+
+        return null;
+    }
+
```
nmetulev, line 155: **M5 (Medium) — `WINAPP_DBGTOOLS_DIR` override causes wasted egress + a false "unavailable".** When the override is set, `CandidateDirectories` `yield break`s here, so only the override dir is ever inspected. But the caller (`XamlTriageService.cs:44-54`), on a null resolve, still runs `TryAcquireFromNuGetAsync(cacheBinDir)` + `WinDbgJsProviderAcquirer.TryAcquireAsync(cacheBinDir)` — downloading ~23 MB into the *cache* dir — then re-resolves against only the override again → returns `null` → `UnavailableNote()`.

**Reproduced live:** override → empty dir with a fully warm cache still reported *"skipped … set the `WINAPP_DBGTOOLS_DIR` environment variable to a debugger directory that contains them"* (the variable that was already set). On a cold cache this is real network egress despite an "authoritative" override.

**Fix:** either keep `cacheBinDir` / installed roots as fallbacks even when an override is set, or skip the acquire step when an (incomplete) override is in force.

nmetulev (commented):
## Review: WinUI stowed-exception triage — findings + live end-to-end verification

I reviewed this across security, correctness, CLI UX, alternative-solution, test-coverage, docs/samples, and packaging, cross-checked the top findings on a second model family (GPT-5.4), and **validated the feature end-to-end** against a purpose-built crashing WinUI 3 app.

### End-to-end verification ✅
Built a packaged WinUI 3 app that auto-crashes ~2 s after activation with `0xC000027B`, then ran the branch-built `winapp run --debug-output` against it:
- **Happy path works:** ClrMD managed stack + full `!xamlstowed`/`!xamltriage` breakdown (`ErrorCode 0xc000027b`, `ErrorMessage "IDispatcherQueueHandler::Invoke failed."`, native `CoreMessagingXP!DispatcherQueueTimer::TimerCallback → RoOriginateError` dispatch stack).
- **Acquisition pipeline validated:** engine bits copied from the NuGet cache (pinned `20260319.1511.0`), **`JsProvider.dll` (~23 MB) downloaded from the WinDbg CDN + Authenticode-verified**, `winui-dbgext.js` (137 KB) from GitHub — all cached under `~\.winapp\dbgtools\`.
- **Warm-cache re-run: 8.7 s**, no re-download.
- **M5 reproduced live** (see inline): with `WINAPP_DBGTOOLS_DIR` pointed at an empty dir and a fully warm cache, triage still reported *"unavailable — set `WINAPP_DBGTOOLS_DIR` …"* (the variable that was already set).

### Findings
| ID | Sev | Location | Issue |
|----|-----|----------|-------|
| H1 | **High** | `CrashDumpService.cs:202` + `XamlTriageService.cs:101` | Triage `OperationCanceledException` (incl. `HttpClient` timeout) escapes and **discards the already-computed managed crash stack** → "Analysis failed." Regression on slow / first-run networks. |
| M1 | Med | `.github/plugin/agents/winapp.agent.md:139` | `--debug-output` agent doc not updated for triage / `WINAPP_DBGTOOLS_DIR` / first-run download. |
| M2 | Med | tests | No integration test asserts triage failure is non-fatal (fail-open) — would have caught H1. |
| M3 | Med | tests | `AuthenticodeVerifier` (the trust gate for a DLL loaded into the debugger) has no unit tests. |
| M4 | Med | `XamlTriageBinaries.cs:287` | Engine `.nupkg` downloaded + extracted with no package-hash/signature verification (HTTPS-only trust). |
| M5 | Med | `XamlTriageBinaries.cs:107` | `WINAPP_DBGTOOLS_DIR` override short-circuit → wasted egress into an ignored cache dir, then false "unavailable." |
| M6 | Med | `WinDbgJsProviderAcquirer.cs:66` | Non-atomic writes to final filenames (23 MB JsProvider + `File.Copy` sites) → concurrent / partial-file races. |
| L1 | Low | `RunCommand.cs` help | `--debug-output`/`--symbols` help text doesn't mention triage or first-run downloads. |
| L2 | Low | `AuthenticodeVerifier.cs:82` | Revocation checking disabled (`WTD_REVOKE_NONE`) — no CRL/OCSP. |
| L3 | Low | `CrashDumpService.cs:234` | "triage written to the debug log" prints even when triage was **skipped/unavailable**; native value-add never surfaced in console. |
| L4 | Low | `XamlTriageBinaries.cs:157` | Cached `JsProvider.dll` not re-verified on cache hit (only at download time). |
| L5 | Low | `XamlTriageBinaries.cs:275` | Bespoke NuGet download duplicates `INugetService` (maintainability — note: `NugetService` also skips integrity checks, so reuse would **not** fix M4). |

### Positives
- `JsProvider.dll` is HTTPS + Authenticode-gated; the extension script is git-blob-SHA1-pinned and fail-closed; `ZipArchive.ExtractToDirectory` avoids zip-slip; the ranged ZIP64 reader uses bounded reads. Packaging is clean (exact-pinned Microsoft packages, restore-only refs, AOT end-user path falls back to flat-container download).

Inline comments below anchor the actionable code-level items (H1, M4, M5, M6, L2, L3). H1 is the one blocking-worthy item; the rest are polish/hardening.

src/winapp-CLI/WinApp.Cli/Services/WinDbgJsProviderAcquirer.cs
```diff
@@ -0,0 +1,227 @@
...
+            var bytes = await ExtractJsProviderAsync(reader, msixName, pathPrefix, cancellationToken);
+            if (bytes == null)
+            {
+                logger.LogDebug("JsProvider.dll not found in WinDbg bundle for {Msix}/{Prefix}.", msixName, pathPrefix);
+                return false;
+            }
+
+            Directory.CreateDirectory(destDir.FullName);
+            var targetPath = Path.Combine(destDir.FullName, TargetFileName);
+            await File.WriteAllBytesAsync(targetPath, bytes, cancellationToken);
```
nmetulev: **M6 (Medium) — non-atomic write to the final filename.** The 23 MB `JsProvider.dll` is written directly to its final cache path here, *then* Authenticode-verified at line 71 (and deleted if invalid). A second concurrent `winapp run --debug-output` — or the same run's later `ResolveExisting` — can observe a partially-written or not-yet-verified DLL at this path and `.load` it. The `File.Copy(..., overwrite: true)` sites in `XamlTriageBinaries.cs` (lines 251, 309) have the same shape.

**Fix:** write to a temp file in the same directory and atomically `File.Move` it into place only after verification succeeds.

src/winapp-CLI/WinApp.Cli/Helpers/AuthenticodeVerifier.cs
```diff
@@ -0,0 +1,145 @@
...
+                pcwszFilePath = pPath,
+                hFile = IntPtr.Zero,
+                pgKnownSubject = IntPtr.Zero,
+            };
+
+            var data = new WINTRUST_DATA
+            {
+                cbStruct = (uint)sizeof(WINTRUST_DATA),
+                dwUIChoice = WTD_UI_NONE,
+                fdwRevocationChecks = WTD_REVOKE_NONE,
```
nmetulev: **L2 (Low) — revocation checking disabled.** `fdwRevocationChecks = WTD_REVOKE_NONE` (and `dwProvFlags = WTD_REVOCATION_CHECK_NONE` at line 86) means a revoked signing certificate would still pass. Since this gate protects a DLL loaded into the debugger process, consider `WTD_REVOKE_WHOLECHAIN` with a best-effort/cached fallback for offline scenarios.

src/winapp-CLI/WinApp.Cli/Services/CrashDumpService.cs
```diff
@@ -166,6 +229,11 @@ await File.AppendAllTextAsync(logPath,
                 console.MarkupLine("[red][[CRASH ANALYSIS]][/]");
                 console.WriteLine(summary);
                 console.MarkupLine("[red]=====================================[/]");
+                if (!string.IsNullOrWhiteSpace(xamlTriage))
+                {
+                    console.MarkupLine("[dim]WinUI stowed-exception triage written to the debug log.[/]");
```
nmetulev: **L3 (Low) — misleading success message.** This prints whenever `xamlTriage` is non-empty — but the skip/unavailable path also returns a non-empty note (`UnavailableNote()` / "WinUI Triage: skipped — …"), so the console shows *"triage written to the debug log"* even when triage was skipped. **Verified live:** with the cache made unreachable, this exact line still printed while the log actually said *"could not be obtained."*

**Consider** distinguishing success vs skip, and surfacing a one-line verdict (`ErrorCode` + `ErrorMessage`) in the console rather than only pointing at the log — right now the headline native value-add never appears in the terminal.

src/winapp-CLI/WinApp.Cli/Services/WinDbgJsProviderAcquirer.cs
```diff
@@ -0,0 +1,233 @@
...
+/// <remarks>
+/// The bundle is a ZIP64 archive whose per-architecture inner <c>.msix</c> packages are STORED
+/// (uncompressed), so the inner archive can be addressed as a contiguous sub-range and parsed in
+/// place. <c>JsProvider.dll</c> inside the inner msix is deflate-compressed and is inflated after
+/// extraction. See <see cref="ZipRangeExtractor"/> for the underlying range-based ZIP reader.
+/// </remarks>
+internal static class WinDbgJsProviderAcquirer
+{
+    // Stable entry point published by the WinDbg team; resolves to the current MainBundle URI.
+    private const string AppInstallerUrl = "https://windbg.download.prss.microsoft.com/dbazure/prod/1-0-0/windbg.appinstaller";
```
nmetulev: **Blocking — this "current" source drifts out of sync with the version-pinned engine.**

`AppInstallerUrl` resolves to the WinDbg *current* MainBundle, so the `JsProvider.dll` extracted here rolls forward independently of the NuGet-pinned `DbgEng` (`XamlTriageBinaries.cs:66`, product `10.0.29547.1002`). Today "current" yields `JsProvider.dll` **10.0.29617.1000** (30.7 MB); loading it into the older engine makes the triage child exit `0x80000003` and triage is silently skipped.

Reproduced from an empty `~\.winapp\dbgtools\` on **both** the dev/CI copy path and the real end-user api.nuget.org download path — a version-matched 23 MB `JsProvider.dll` (10.0.29547.1002) is the only thing that makes triage succeed. `FallbackBundleUrl` (line 32) is only used on appinstaller *parse failure*, so the matched bundle is never the primary source.

**Fix:** make a version-matched WinDbg bundle the primary source (pin it), or add a post-acquire `dbgeng`/`JsProvider` `ProductVersion` compatibility check that rejects a mismatched provider (re-acquire or fail-closed).

src/winapp-CLI/WinApp.Cli/Services/XamlTriageBinaries.cs
```diff
@@ -0,0 +1,488 @@
...
+        !string.IsNullOrWhiteSpace(Environment.GetEnvironmentVariable(EnvOverride));
+
+    private const string FlatContainer = "https://api.nuget.org/v3-flatcontainer";
+
+    /// <summary>
+    /// Pinned native-debugger package version. Must stay in sync with the
+    /// <c>Microsoft.Debugging.Platform.*</c> entries in <c>Directory.Packages.props</c>; a unit test
+    /// asserts they match so runtime acquisition uses the same version that <c>dotnet restore</c> pins.
+    /// </summary>
+    public const string DbgPackageVersion = "20260319.1511.0";
```
nmetulev, line 67: The engine is exactly version-pinned **and** SHA-512-verified here (nice), but `JsProvider.dll` — which is `.load`ed into the same engine — is acquired from an unpinned "current" WinDbg bundle (`WinDbgJsProviderAcquirer.cs:29`), and **nothing asserts the two builds agree**. The `.nupkg` SHA-512 drift test you added is the right pattern; consider an analogous assertion/enforcement that the acquired `JsProvider` build matches this pinned engine build so a fresh-cache user can't end up with an incompatible pair.

nmetulev (commented):
## Follow-up review: all prior findings resolved — but stress-testing the real "download on first use" path surfaced a blocking issue

Thanks for the thorough pass on `11cafc7` — **every finding from my first review is addressed**, several with extra hardening (SHA-512 per-`.nupkg`, `AtomicFile`, JsProvider re-verify self-heal, revocation whole-chain). Re-verification table is at the bottom.

I then did what I'd flagged I hadn't done live the first time: **exercised the genuine cold-cache "download on first use" path** (my earlier happy-path run reused a warm cache from an earlier session). That's where the problem is.

### 🔴 Blocking — `JsProvider.dll` / engine **version drift** breaks triage for every fresh user

The engine and the JS provider are pinned to **different, incompatible sources**:

| Binary | Source | Version I got today |
|--------|--------|--------------------|
| `dbgeng.dll` (engine) | NuGet `Microsoft.Debugging.Platform.DbgEng` `20260319.1511.0`, SHA-512-gated | **10.0.29547.1002** |
| `JsProvider.dll` | WinDbg **"current"** appinstaller (`WinDbgJsProviderAcquirer.cs:29`) | **10.0.29617.1000** (30.7 MB) |

The "current" WinDbg bundle rolls forward independently of the NuGet-pinned engine. Loading the newer JS provider into the older engine makes the triage child process exit `0x80000003` (STATUS_BREAKPOINT, `-2147483645`), and triage is **silently skipped** — the exact headline feature this PR adds never runs.

**Reproduced from a genuinely empty `~\.winapp\dbgtools\`, both acquisition paths:**

- **Scenario A — dev/CI cold** (`NUGET_PACKAGES` has the debugger packages): engine **copied** from the NuGet cache, JsProvider **downloaded** (30.7 MB "current") → **triage skipped**, ~16.6 s.
- **Scenario B — true end-user cold** (`NUGET_PACKAGES` empty → engine must download): engine **downloaded from api.nuget.org flat-container** (the live `.nupkg` **passed the compiled SHA-512 gate** ✅ — that path is solid), JsProvider **downloaded** (30.7 MB "current") → **triage skipped**, ~20.7 s.
- **Control:** dropping in a version-matched `JsProvider.dll` (10.0.29547.1002, 23 MB) → triage **succeeds** with the full verdict (`0xc000027b — "IDispatcherQueueHandler::Invoke failed."`).

So the *only* variable separating success from skip is the JsProvider build. **Why it hides in dev:** iterative dev keeps a warm cache whose JsProvider was fetched when "current" happened to equal the pinned engine build — which is exactly why my first-pass happy path looked green. A user (or CI, or me with a cleared cache) downloading today gets the mismatch.

**Root cause:** `AppInstallerUrl` always resolves to the "current" MainBundle. `FallbackBundleUrl` (`WinDbgJsProviderAcquirer.cs:32`) is only used when the appinstaller *fails to parse*, so a version-matched bundle is never the primary source, and nothing checks that the two binaries agree after acquisition.

**Any one of these closes it:**
1. Pin JsProvider to a WinDbg bundle whose build matches the pinned `DbgEng`, and make that bundle the **primary** source (not just the parse-failure fallback). Add a unit test asserting `JsProvider` build == engine build — mirroring the excellent `.nupkg` SHA-512 drift test you already added for the engine.
2. **Or** add a post-acquire compatibility check: compare `dbgeng.dll` and `JsProvider.dll` `ProductVersion`; on mismatch, reject the JsProvider (re-acquire the matched bundle, or fail-closed) so self-heal + honest messaging kick in instead of a silent child crash.
3. Either way, surface a clearer skip reason than `exited with code -2147483645` — detect the mismatch and say so.

### Residuals (non-blocking)

- **`dbgeng.dll` corruption has no self-heal (Low).** `ResolveExisting` re-verifies only `JsProvider.dll`'s Authenticode on a cache hit; a truncated/corrupt engine DLL passes `File.Exists`, the child crashes, triage skips, and the cache stays poisoned across runs (JsProvider self-heals, engine doesn't). Managed stack is still preserved. Consider extending the re-verify (size or hash) to the engine bits too.
- **H1 console noise on the download timeout (Low).** The H1 fix correctly **preserves the managed stack** now (verified with a live 5-min induced timeout 👍), but the raw `TaskCanceledException` stack still prints to the console and the user waits the full 5-minute `HttpClient.Timeout` before the crash stack appears. Consider a shorter default JsProvider-acquisition timeout and suppressing the raw stack.

### Re-verified fixed this pass ✅
H1 (managed stack preserved on timeout — double-guarded) · L3 (`XamlTriageResult` surfaces the one-line verdict in console) · M5 (override no longer wastes egress; `DescribeOverrideGap` is honest) · M6 (`AtomicFile` staged-write + publish) · L4 (cached JsProvider re-verified every resolve → truncation self-heals) · M4 (compiled-in SHA-512 per `.nupkg`, `latest` fallback dropped → fail-closed; drift test passes) · L2 (revocation whole-chain, cache-only so no offline hang; `CERT_E_REVOKED` hard-fails). Triage unit tests: **63/63 pass** (the 5 failing `E2E_NodeSubcommand_*` are pre-existing and unrelated — they need `npm run build`).

nmetulev:
✅ **Revalidated `3711ec2` end-to-end — the version-drift fix works, and both residuals are closed.**

Rebuilt at `3711ec2` and ran the triage unit suite (**43/43 pass**, including the new `VersionsMatch` cases and the `PinnedJsProviderProductVersion` drift guard), then re-ran the cold-cache scenarios that were failing before:

**Cold first-run (the scenario that was broken):** cleared `~\.winapp\dbgtools\` and forced a fresh acquire → `JsProvider.dll` now downloads from the pinned bundle as **22.3 MB / `10.0.29547.1002`** (matches the engine) → **triage SUCCEEDS** with the verdict `0xc000027b — "IDispatcherQueueHandler::Invoke failed."`, ~18.6 s. (Before this commit: 30.7 MB / `10.0.29617.1000` → silently skipped.) 🎯

**Compat-gate self-heal:** planted a valid Microsoft-Authenticode-signed but wrong-version DLL (`14.44.35222.100`) as `JsProvider.dll` → on the next run it was rejected by the version gate, the matched provider was re-downloaded, and triage succeeded (cache healed back to `10.0.29547.1002`).

**Engine self-heal (my earlier residual):** truncated the cached `dbgeng.dll` to 2 KB → `IsUsablePeFile` + the unreadable-engine-version compat path rejected it, the engine was re-acquired, and triage succeeded.

Nice work — you took both recommendations (pin the matched bundle **and** add a runtime compat gate), closed the two residuals I'd left (engine PE self-heal + the H1 timeout console noise via the new `OperationCanceledException` fail-open), and the `0x80000003` exit now explains itself. From my side this resolves the blocking finding — no further concerns.

One tiny non-blocking thought: `PinnedJsProviderProductVersion_MatchesEngineDrift_Guard` asserts against a hardcoded `"10.0.29547.1002"` rather than reading `dbgeng.dll`'s actual product version, so a future `DbgPackageVersion` bump that updates neither constant wouldn't be caught by that specific test. The runtime compat gate still fails closed regardless, so it's belt-and-suspenders either way.

nmetulev (approved):

PR #601, opened 2026-07-06
Add ui audit command for per-view accessibility checks

yeelam-gordon:
## Summary
- add `winapp ui audit` for names, keyboard, screen-reader-readiness, contrast, and role checks
- support `--area` and `--level` to scope the audit and choose AA/basic vs AAA/thorough depth
- make the command description explicit that it audits the **currently visible view** and should be paired with other UI automation to move through additional pages/tabs/states
- regenerate the CLI schema and UI automation skill/docs

## Why
This adds a high-signal accessibility/contrast lint over the current UI view while keeping whole-app coverage composable with the rest of the UI automation primitives.

### DIFF
--- a/.claude/skills/winapp-ui-automation/SKILL.md
+++ b/.claude/skills/winapp-ui-automation/SKILL.md
@@ -200,6 +200,27 @@ winapp ui wait-for btn-submit-a1b2 -a myapp --timeout 5000
 winapp ui wait-for itm-status-c3d4 -a myapp --value "Complete" --timeout 5000
 ```
 
+### Audit accessibility & contrast
+```powershell
+# Audit the whole window (all areas, basic level) — exits non-zero if any FAIL is found (CI gate)
+winapp ui audit -a myapp
+
+# Machine-readable report ({ summary, issues }); write it to a file too
+winapp ui audit -a myapp --json -o audit.json
+
+# Scope to specific areas (repeatable). Areas: names, keyboard, screen-reader, contrast, roles
+winapp ui audit -a myapp --area names --area keyboard
+
+# Go deeper: 'thorough' adds heuristic rules (e.g. tab-order coherence) and applies WCAG AAA contrast
+winapp ui audit -a myapp --level thorough
+
+# Contrast only, at the default basic level (WCAG AA thresholds)
+winapp ui audit -a myapp --area contrast
+```
+- `--area` selects one or more audit areas (default: all). Contrast requires a window pixel capture; it is measured only when the `contrast` area is selected.
+- `--level basic` (default) runs fast essential rules with WCAG **AA** contrast thresholds (normal 4.5, large 3.0); `--level thorough` adds heuristic/deeper rules (e.g. keyboard tab-order) and applies WCAG **AAA** contrast thresholds (normal 7.0, large 4.5).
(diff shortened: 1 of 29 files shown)


nmetulev (commented):

**Verdict: 🟡 SHIP-WITH-CHANGES** — Sound architecture and a genuinely useful idea, but two correctness bugs make the current output misleading, and there's a scope question worth a deliberate decision.

### What I actually ran
Built (~41s, 0 warnings); **23 engine + 13 orchestrator + 126 ui** tests pass. Audited Notepad and Calculator:
- `--area` scoping, `--level basic/thorough`, `--json`, exit codes (0 pass / 1 fail): all work; the `{summary, issues}` JSON envelope is stable.
- Contrast AA on Notepad's status bar returned **PASS at a measured ~3.15–3.34:1** — a false negative (see H1).
- `--level thorough` (AAA) correctly tightened thresholds and flagged more.
- Calculator returned **0 elements** but printed "✓ No accessibility issues found" (M5 — indistinguishable from genuinely clean).

### Must-fix (High)
- **H1 — DPI-dependent large-text false negatives.** "Large text" is decided by bounding-box height (`LargeTextHeightPx = 24`, `UiAuditEngine.cs:91`), not font size in points. At 125–150% DPI, normal ~12pt text is classified "large" and judged against the lenient 3.0:1 threshold, so ~3.2:1 text **passes AA**. Real false negatives on the majority of displays; the unit tests use a fixed height of 16 and miss it. Read font size from `TextPattern` (`UIA_FontSizeAttributeId`), or normalize the pixel height by the window DPI scale.
- **H2 — Pane false positives.** `Pane + IsInvokable` is flagged "interactive but has no accessible name" (`UiAuditEngine.cs:128-136`). Shell/framework panes expose `InvokePattern` structurally, producing name-FAIL + role-WARN noise on nearly every app (confirmed on Notepad). Exclude container types (Pane/Group/Window) unless also keyboard-focusable.

### Medium / Low (selected)
- **M1** contrast epsilon `r + 0.05 < threshold` silently passes ratios up to 0.05 below AA (e.g. `#777` on white = 4.478:1 not flagged). Use `r < threshold`.
- **M2** the `screen-reader` area is a static UIA proxy; the name over-promises AT-driven testing.
- **M3** the pass counter sums per-check hits ("103 passed" for ~30 elements).
- **M4** `AutomationId == Name` WARN false-positives on standard Win32 menu items (File/Edit/View).
- **M5** 0-element result is reported as clean.
- **L1** sRGB linearization uses the WCAG 2.1-errata `0.03928` vs the corrected `0.04045`. **L3** `--level aa`/`aaa` aliases are rejected (only `basic`/`thorough`).

### Fit & recommendation
This is the largest of the five PRs and the one that moves furthest from packaging/automation into **accessibility certification** — a space owned by mature dedicated tools (Accessibility Insights for Windows, Arc, Axe). The value-add here is the CLI/agent-composable *interface*, not the checks themselves. The risk: an incomplete or incorrect auditor gives **false confidence** for Store submission. If it ships, the DPI + contrast bugs are non-negotiable, and the docs should state plainly that it's a **quick lint, not a WCAG cert**. I'd treat inclusion as a deliberate scope decision rather than an automatic yes.

chiaramooney:
I believe Accessibility Insights has "Fast Pass" which runs this same set of checks on whatever page of an app you point it to. It would be good to investigate if we can reuse Fast Pass here in some way so we don't have to maintain our own dictionary of accessibility rules and can lean on accessibility insights to update and maintain the correct accessibility policies.

PR #602, opened 2026-07-06
Add ui watch command for UI event streaming

yeelam-gordon:
## Summary
- add `winapp ui watch` to listen for UIA/WinEvent notifications from a running app
- support event filtering, bounded capture (`--duration-sec`, `--max-events`), and JSON/NDJSON output
- regenerate the CLI schema and UI automation skill/docs

## Why
This adds a first-class way to observe focus, selection, window, and related UI events instead of polling for every state transition.


nmetulev (commented):

**Verdict: 🟡 SHIP-WITH-CHANGES** — Good architecture (dedicated STA pump, correct COM handler teardown), but a buffering bug currently defeats the streaming use case this command exists for.

### What I actually ran
Built (~33s, 0 warnings); 17 unit tests pass. Watched Notepad while driving it from a second shell:
- `--duration-sec` and `--max-events` bounding stop cleanly; NDJSON is one event per line + a summary object. ✅
- Reading the `-o` file **mid-run: empty** until the process exits (C01) — real-time streaming doesn't work.
- Opening one WinUI3 Notepad emitted **25+ `window-open` events** for internal panes (`DesktopWindowXamlSource`, `Default IME`, `CicMarshalWnd`, …); only 1 was the real window (H01).
- Focus-event latency <100ms; timestamps ISO-8601 UTC.

### Must-fix (Critical / High)
- **C01 — no flush after each event** (`UiWatchCommand.cs:173`). `StreamWriter` defaults to `AutoFlush = false`, so neither stdout nor the `-o` file streams until the process exits. This is the headline "agents read events live" scenario, and it's currently broken. One-line fix (`Flush()` after each write).
- **H01 — window-open/close filter too broad** (`UiEventWatcher.cs:225-230`): 20–30 internal-pane events per window open. Add a top-level-window filter (parent == desktop, or `WS_CAPTION`/`WS_OVERLAPPEDWINDOW`).
- **H02 — event `selector` isn't a winapp slug.** It's the raw AutomationId (`UiEventWatcher.cs:314-319`) and is frequently null, so `event.element.selector` can't be piped into `ui invoke`/`set-value`. The unit-test fakes use slug-format strings, which hides this. Route through `SelectorService`.

### Medium / Low
- **M01** `RequestStop()` is a non-atomic check-then-set. **M02** the `s_active` static is unguarded against a concurrent `WatchAsync`. **M03** `--output` description is copy-pasted from `screenshot`. **M04** npm `winapp-commands.ts` is missing `uiWatch()`.
- **L**: no integration test for the real watcher (all 17 use `FakeUiEventWatcher`); the summary line lacks a `type` discriminant; `ui-json-envelope.md` isn't updated; the `detail` field format varies across event types.

### Fit & recommendation
Push notifications are genuinely more efficient than polling. But this is a long-running **streaming** command inside an otherwise request/response CLI, and it overlaps the existing `wait-for` verb. The agent workflow (fork a subprocess, drive the app from another, parse after exit) is awkward, and the flush bug amplifies it. I'd gate this on fixing streaming (**C01**), then decide whether the streaming model earns its complexity versus investing in `wait-for`.

PR #603, opened 2026-07-06
Add --allow-system-keys to ui send-keys

yeelam-gordon:
## Summary
- add `--allow-system-keys` to `winapp ui send-keys`
- keep the default safety posture (system-/shell-reserved combos are still blocked) and require an explicit opt-in for OS-wide hotkeys via `--via send-input`
- preserve the existing no-target, foreground, and secure-desktop safeguards

## Why
This enables intentional global hotkey automation (for example, PowerToys-style Win-key shortcuts) without weakening the default behavior.


nmetulev (commented):

**Verdict: 🟡 SHIP-WITH-CHANGES** — The double opt-in design is sound and the existing guards stay intact, but `win+l` should remain blocked and the silent no-op path needs a warning.

### What I actually ran
Built (~33s, 0 warnings); **55/55** SendKeys + SystemKeyGuard tests pass. Against Notepad:
- Normal text `send-keys` ✅.
- `win+l` via `--via send-input` **without** the flag → **BLOCKED**, exit 1, with a clear, actionable error naming the combo and "Pass `--allow-system-keys` to opt in." ✅
- `win+up` **with** `--allow-system-keys --via send-input` → fired OS-wide (Notepad maximized); `win+down` restored it — self-reversing ✅. (Injection works here when the target is foregrounded first.)
- `win+l` **with** `--allow-system-keys` but **no** `--via send-input` → exit 0, `via":"post-message"`, no warning, hotkey never fires (SEC-02).
- No-target guard still fires with the flag ✅.

### Must-fix
- **SEC-01 (High) — `win+l` is unblocked by the flag** with no per-combo granularity (`SystemKeyGuard.cs:56-59` collapses everything to a generic `win+<key>`; bypass at `UiSendKeysCommand.cs:244`). `win+l` is **not** a SAS — a user-mode `SendInput` of VK_LWIN+L reaches explorer's low-level hook and calls `LockWorkStation()`. A mistyped `win+l` vs `win+left` in automation locks the (remote) desktop. Consider keeping `win+l` (and `alt+f4` / `ctrl+shift+esc`) blocked **even with** the opt-in; the cited use case is registered hotkeys like `win+shift+v`, not lock control.
- **SEC-02 (Medium) — silent no-op.** `--allow-system-keys` without `--via send-input` is silently accepted (exit 0, no effect). Emit a `LogWarning` so this isn't a silent footgun.
- **COR-01** add a test for the `--allow-system-keys` + `post-message` path. **DOCS-02** the docs name only `ctrl+alt+del` as "still blocked," implying other combos are safe — call out `win+l` / `alt+f4` explicitly.

### Fit & recommendation
Real, unmet use case (global hotkeys are unreachable via `post-message`), default posture unchanged, guards preserved, an actionable agent-readable error, and a `LogWarning` audit trail on every bypass. The tension is the security surface: a CLI/agent can now fire OS-wide hotkeys. The blanket on/off is acceptable to start **if** the docs and the `win+l` gap are addressed; a per-combo allowlist (e.g. `--allow-system-keys win+shift+v`) would be a cleaner long-term design and would keep the blast radius to what the caller actually needs.

nmetulev (approved):

PR #618, opened 2026-07-12
[feat]: Add #winapp/bindings package import for JS bindings

lei9444:
## Description

Adds a stable `#winapp/bindings` [Node package import](https://nodejs.org/api/packages.html#subpath-imports) as the recommended way to load JS bindings from Electron / Node projects, replacing brittle relative paths like `require('../.winapp/bindings/index.js')`.

**Why**

• Relative paths break when files move or when bindings live under a workspace root that differs from the consumer file.
• A single well-known specifier makes docs, samples, and guides copy-paste friendly regardless of where user code sits.
• Node's subpath imports are the standard mechanism for this and work in both CJS and ESM.

## Usage Example

After `npx winapp init`, `package.json` gets:

```
{ "imports": {
    "#winapp/bindings": "./.winapp/bindings/index.js",
    "#winapp/bindings/*": "./.winapp/bindings/*" }
}
```

App code becomes:

`const { Windows } = require('#winapp/bindings');`

## Type of Change

- ✨ New feature

## Checklist

- [x] [Language-specific guides](../docs/guides) updated (if applicable)
- [x] [Sample projects updated](../samples) to reflect changes (if applicable)

## AI Description

This pull request introduces a stable import for the JavaScript bindings using `#winapp/bindings`, replacing previously used relative paths. This change makes it easier to load bindings in Electron and Node projects by standardizing the import method, which enhances documentation and usability. Users can now import bindings as follows:
```js
const { Windows } = require('#winapp/bindings');
```
A prerequisite for this import is having `@microsoft/dynwinrt-codegen` version `≥ 0.1.0-preview.8`.


yeelam-gordon (approved):

nmetulev (commented):
## 🤖 AI-generated review

> This review was generated by **GitHub Copilot CLI** (multi-dimensional PR-review skill: security, correctness, CLI UX, alternative-solution, test-coverage, docs/samples, packaging, plus a GPT-5.5 cross-check). Findings were **verified by actually running the code**, not just reading it. Please treat it as assistive input and apply your own judgement.

**Scope:** `leilzh/useimports` vs `origin/main` — 3 commits, 20 files, +460/-59.
**Verdict:** 1 High · 3 Medium · 2 Low. Nothing blocks merge except **H1** (two-line fix). Security, correctness, docs/samples, packaging, and alternative-solution dimensions came back clean; the electron-forge regex and the shipped sample config were verified safe.

---

### 🔴 H1 (High) — import hints leak to stdout under `--quiet` / `--json`

**`src/winapp-npm/src/jsbindings/cli-hooks.ts:402-407`** · confirmed by GPT-5.5 cross-check · **empirically reproduced**

The new `#winapp/bindings` hints are emitted with **no `!quiet` guard**, unlike every sibling hint (L359, L414, L427). Under `--quiet`/`--json`, `group` is `undefined`, so `bufferingLog` falls back to `taskLog` (L346), which calls `console.log` unconditionally (L56-68). This corrupts the machine-readable output that `--json` explicitly promises to keep clean ("_so we never corrupt downstream parsers_", L249).

Reproduced by mimicking `handleInit`'s exact `--json` wiring with the real compiled functions — both lines hit stdout:

```
  💡 Added "#winapp/bindings" package imports to package.json.
  ⚠ Existing "#winapp/bindings" alias in package.json differs from the winapp default; leaving as-is. ...
```

Reachable via `winapp init --add-js-bindings --json` with codegen ≥ preview.8 (now the shipped sample version).

**Fix** — keep the mutation, guard only the emission:
```ts
if (supportsPackageImports(codegenVersion)) {
  const result = ensureJsBindingsImports(workspaceDir);
  if (!quiet) emitJsBindingsImportsHints(result, bufferingLog);
}
```

---

### 🟡 Medium

- **M1 — silent skip has no guidance** · `cli-hooks.ts:387-407` (cli-ux)
  When installed codegen is older than preview.8, `init --add-js-bindings` silently skips the imports map — but the guides tell users to `require('#winapp/bindings')`, so they hit an unexplained resolution failure. Add an `else` hint naming the cause + the upgrade command (`npm i -D @microsoft/dynwinrt-codegen@latest && npx winapp init --add-js-bindings`), matching the fallback wording already in `index.md`.

- **M2 — init version-gate / quiet path untested** · `cli-hooks.ts:387-407` (test-coverage)
  The new gate + config-only imports write has no test; a `--json` assertion would have caught **H1**. Add a `handleInit` test (preview.7/null → imports not written; preview.8 → written) and a `--quiet`/`--json` no-leak case.

- **M3 — `getCodegenPackageVersion` untested** · `codegen-runner.ts:488-492` (test-coverage)
  New function that drives the gate; `JSON.parse` can throw. Cover valid version, non-string/missing `version` → `null`, and malformed/missing `package.json` → throws.

### 🟢 Low (behavior verified correct — regression-protection only)

- **L1** — `supportsPackageImports` lacks cases for `0.1.0-preview.10`, `0.1.1-preview.1`, `1.0.0`, `0.1.0+build`. I ran all four against the real function — all correct. `codegen-version.test.ts:9-22`
- **L2** — `ensureJsBindingsImports` only tests `imports: 'invalid'`. I verified `null`/`[]` → throw and `{}` → adds both aliases. Add these cases. `package-json-config.test.ts:154-159`

---

<details>
<summary>✅ What was verified by running it</summary>

| Check | Result |
|-------|--------|
| H1 leak under `--json` (real compiled fns, captured stdout) | **Leaked** — both hint lines reached stdout |
| electron-forge asar ignore regex vs 8 sample paths | Keeps `.winapp/bindings/**`, ignores `.winapp/include\|generated` — safe |
| `samples/electron/forge.config.js` | Already uses the new regex (not in diff) — no packaging break |
| `supportsPackageImports` over 18 versions | Correct (`preview.10`→T, `0.1.1-preview.1`→T, `1.0.0`→T, `beta.1`→F) |
| `ensureJsBindingsImports` null/`[]`/`{}` | Correct (throw / throw / adds aliases) |
| `npm test` baseline | **192/192 pass** |

</details>

<sub>🤖 Generated by GitHub Copilot CLI · verified against the checked-out branch · not a substitute for human review.</sub>

nmetulev:
> 🤖 **AI-generated follow-up** (GitHub Copilot CLI). Posted by a human-run agent; please sanity-check before relying on it.

### Fixes applied + re-reviewed → 0 remaining issues

Following the [AI review above](#), the findings have been addressed on this branch and I re-ran the same multi-dimension review (security, correctness, CLI-UX, alternative-solution, test-coverage, docs/samples, packaging + a cross-model check on a different model family) against the updated code.

**What changed**
- **H1 (high · UX):** `#winapp/bindings` import hints no longer leak to stdout under `--quiet` / `--json`. Hint emission now routes through a pure `planJsBindingsImportsHints(input, { quiet })` gate that returns `[]` when quiet. The `package.json` `imports` mutation still runs under quiet — only the hint *text* is suppressed.
- **M1 (medium · UX):** when the installed `@microsoft/dynwinrt-codegen` is too old (< preview.8), init now prints an actionable upgrade hint (upgrade command + relative-path fallback) instead of skipping silently.
- **M2 / M3 / L1 / L2 (test-coverage):** added tests for the quiet/version gate, `getCodegenPackageVersion`, `supportsPackageImports` edge cases (`preview.10`, `0.1.1-preview.1`, `1.0.0`, `+build`), and `ensureJsBindingsImports` (null / array / empty imports).
- Removed the now-orphaned `emitJsBindingsImportsHints` helper left behind by the refactor.

**Re-review result:** 0 critical / 0 high / 0 medium / 0 low. The cross-model check independently confirmed the quiet-leak is resolved and the "mutation still runs under quiet" invariant holds.

**Verified by running it** (`src/winapp-npm`): `npm test` → **203/203 pass**, `eslint` clean, `prettier --check` clean. Also exercised the compiled gate directly: quiet ⇒ no lines (both the configured and unsupported branches); non-quiet ⇒ correct add / divergence-warning / upgrade hints.

PR #621, opened 2026-07-13
Fix @-prefixed args swallowed by response-file expansion (#619)

### LINKED ISSUE #619: [Bug]: Arguments starting with "@" are swallowed by response-file expansion (e.g. winapp ui search "@assistant")
### Describe the bug

Any command-line argument **value** that starts with `@` is intercepted by System.CommandLine's response-file token expansion instead of being passed through literally. The parser treats `@assistant` as "read arguments from a file named `assistant`", fails at parse time with `Response file not found 'assistant'.`, prints the help text, and exits 1 — the command handler never runs.

This breaks common inputs such as:
- `winapp ui search "@assistant"` / `winapp ui search "@handle"` (chat/social mentions)
- `winapp ui set-value <sel> "@somebody"`
- any positional value that legitimately begins with `@` (emails passed positionally, npm scopes like `@scope/pkg`, C# verbatim identifiers, etc.)

`winapp ui` is used heavily to drive chat/assistant UIs where `@mention` text is extremely common, so this is a frequent foot-gun.

**Root cause:** `WinAppParserConfiguration.Default` (`src/winapp-CLI/WinApp.Cli/Helpers/WinAppParserConfiguration.cs`) sets `EnablePosixBundling = false` but leaves `ResponseFileTokenReplacer` at its default (non-null), so `@`-token expansion stays enabled for every winapp parse.

**Secondary issue:** because the failure happens at parse time, `--json` is ignored — the error prints as plain text + help rather than the standard JSON error envelope, so scripted/LLM callers can't detect it cleanly either.

### To Reproduce

Empirically verified against a Debug build of `src/winapp-CLI` (System.CommandLine 2.0.5):

```
> winapp ui search "@assistant" -a fakeapp --json
  ...help text...
  Response file not found 'assistant'.
  (exit 1)

> winapp ui search "@handle" -a fakeapp --json
  Response file not found 'handle'.
  (exit 1)
```

Control — the same command without a leading `@` reaches the handler normally:

```
> winapp ui search "assistant" -a fakeapp --json
  { "error": { "code": "internal_error", "message": "No running app found matching 'fakeapp'.", ... } }
  (exit 1 — expected, because 'fakeapp' does not exist)
```

### Expected behavior

`@`-prefixed values should be passed through to the command as literal text — `winapp ui search "@assistant"` should search for the literal string `@assistant`. winapp is an app-automation CLI and does not need `@responsefile` indirection, so response-file expansion should be disabled globally.

**Suggested fix:** in `WinAppParserConfiguration.Default`, set `ResponseFileTokenReplacer = null` (property confirmed present on `System.CommandLine.ParserConfiguration` in 2.0.5). Add a regression test (e.g. alongside `OptionTypoValidatorTests`) asserting a `@`-leading argument reaches the handler as a literal.

### Screenshots

Reported from user feedback while driving a chat UI: "@handle assertions fail because `winapp ui search "@assistant"` hits the CLI's own response-file argument parsing."

### OS Version and details

Windows 11 (reproduced on a win-arm64 Debug build). Not OS-specific — it is purely parser configuration.

nmetulev:
## Summary

Fixes #619.

Argument **values** beginning with `@` (e.g. `winapp ui search "@assistant"`) were intercepted by System.CommandLine's response-file token expansion. The parser treated `@assistant` as "read arguments from a file named `assistant`", failed at parse time with `Response file not found 'assistant'.`, printed help, and exited 1 — the command handler never ran. Because the failure happened at parse time, `--json` was also ignored (plain-text error + help instead of the standard JSON error envelope).

This broke common inputs: chat/social `@mention` text (`winapp ui search "@assistant"`, `winapp ui set-value <sel> "@somebody"`), npm scopes (`@scope/pkg`), positional emails, etc. — a frequent foot-gun since `winapp ui` is used heavily to drive chat/assistant UIs.

## Root cause

`WinAppParserConfiguration.Default` set `EnablePosixBundling = false` but left `ResponseFileTokenReplacer` at its (non-null) default, so `@`-token expansion stayed enabled for every winapp parse.

## Fix

Set `ResponseFileTokenReplacer = null` in `WinAppParserConfiguration.Default`, disabling response-file expansion globally so `@`-prefixed values pass through as literal text. winapp is an app-automation CLI and does not need `@responsefile` indirection. All real System.CommandLine parse entry points route through this shared config (`Program.cs` ×2, `CompleteCommand.cs`), so a single change fixes it everywhere — including the secondary `--json` symptom (the handler now runs and emits the normal JSON envelope).

## Trade-offs

The only thing given up is System.CommandLine's `@responsefile` indirection (`winapp <cmd> @args.txt`). It is **undocumented, unused (0 references across CLI source, docs, samples, scripts, npm wrapper, and tests), and ill-suited** to a short-command automation CLI. The bug it fixes has **no workaround today** — with expansion on there is no escape syntax, so `@assistant` is simply broken. The change is trivially reversible if a genuine need ever surfaces.

Rejected alternatives: a replacer that expands only when the file exists (non-deterministic per-cwd) and a `@@file` escaping convention (extra surface for an unused feature) — both worse.

## Tests

New `WinApp.Cli.Tests/WinAppParserConfigurationTests.cs` asserts `@`-leading values parse with no errors and reach handlers as literals:
- `ui search "@assistant"` → `SelectorArgument == "@assistant"`
- `ui set-value txt-name "@somebody"` → `ValueArgument == "@somebody"`
- `ui search "@scope/pkg"` → literal passthrough
- plain-value guard (`assistant`) still parses unchanged

## Verification

- New tests 4/4; existing typo tests 7/7; broader parsing suite (CompleteCommand, WindowsCommandLine, UiCommand, CliSchema, CustomHelp, GlobalOptionPreScan) 199/199.
- Full solution compiles clean in Release (0 warnings / 0 errors).
- **End-to-end** against the built CLI: `winapp ui search "@assistant" -a fakeapp --json` and `"@handle"` now return the standard JSON error envelope (`No running app found matching 'fakeapp'.`) — identical to the plain-`assistant` control — instead of "Response file not found".
- Confirmed no command-surface change (regenerating docs produced only version-string churn, reverted).

No docs/samples changes needed — the only "response file" doc reference is the unrelated cppwinrt `.cppwinrt.rsp` build file.

### DIFF
--- a/src/winapp-CLI/WinApp.Cli.Tests/WinAppParserConfigurationTests.cs
+++ b/src/winapp-CLI/WinApp.Cli.Tests/WinAppParserConfigurationTests.cs
@@ -1,0 +1,74 @@
+// Copyright (c) Microsoft Corporation and Contributors. All rights reserved.
+// Licensed under the MIT License.
+
+using WinApp.Cli.Commands;
+using WinApp.Cli.Helpers;
+
+namespace WinApp.Cli.Tests;
+
+/// <summary>
+/// Regression tests for <see cref="WinAppParserConfiguration.Default"/>.
+/// </summary>
+/// <remarks>
+/// Guards issue #619: argument values that begin with <c>@</c> (chat/social mentions such as
+/// <c>@assistant</c>, npm scopes such as <c>@scope/pkg</c>, positional emails, etc.) must be passed
+/// through to the command handler as literal text and must NOT be interpreted as
+/// System.CommandLine <c>@responsefile</c> indirection. Before the fix, response-file token expansion
+/// was left enabled, so <c>winapp ui search "@assistant"</c> failed at parse time with
+/// "Response file not found 'assistant'." and the handler never ran.
+/// </remarks>
+[TestClass]
+public class WinAppParserConfigurationTests : BaseCommandTests
+{
+    private static string ErrorSummary(System.CommandLine.ParseResult parseResult) =>
+        "Leading '@' value must not trigger response-file expansion. Parse errors: " +
+        string.Join("; ", parseResult.Errors.Select(e => e.Message));
+
+    [TestMethod]
+    public void PassesLeadingAtValueThroughAsLiteral_ForSearchSelector()
+    {
+        var rootCommand = GetRequiredService<WinAppRootCommand>();
+        var args = new[] { "ui", "search", "@assistant", "-a", "fakeapp" };
+
+        var parseResult = rootCommand.Parse(args, WinAppParserConfiguration.Default);
+
+        Assert.AreEqual(0, parseResult.Errors.Count, ErrorSummary(parseResult));
+        Assert.AreEqual("@assistant", parseResult.GetValue(SharedUiOptions.SelectorArgument));
+    }
+
+    [TestMethod]
+    public void PassesLeadingAtValueThroughAsLiteral_ForSetValueArgument()
+    {
+        var rootCommand = GetRequiredService<WinAppRootCommand>();
+        var args = new[] { "ui", "set-value", "txt-name", "@somebody", "-a", "fakeapp" };
+
+        var parseResult = rootCommand.Parse(args, WinAppParserConfiguration.Default);
+
+        Assert.AreEqual(0, parseResult.Errors.Count, ErrorSummary(parseResult));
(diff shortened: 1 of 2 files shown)


zateutsch (approved):
PR review was clean except for a low priority alt solution rec. Approving.

PR #681, opened 2026-07-21
feat: add winapp find-ui — WinUI control & sample search (Gallery + Toolkit + Reactor)

### LINKED ISSUE #651: [Feature]: Port `winui-search` (lexical UI control/sample search) into `winapp`
## Is your feature request related to a problem? Please describe.

There is no first-class way to search WinUI controls and samples from the `winapp` CLI. That capability lives in a separate **`winui-search`** tool in **microsoft/win-dev-skills** (`src/tools/winui-search/`), so it's split across two repos and isn't available to `winapp` users or agents. A developer (or agent) building a Windows UI has to leave `winapp` to discover which control fits and find a working example.

## Current state

- **`winui-search`** (microsoft/win-dev-skills, `src/tools/winui-search/`) does lexical/fuzzy search over WinUI 3 Gallery + Windows Community Toolkit scenarios and controls. It works, but lives outside `winapp`.
- `winapp` has no control/sample search command today.

## Describe the solution you'd like

Bring the `winui-search` capability into `winapp` as a first-class command — `winapp find-ui "<query>"` — so users and agents can search WinUI controls and samples and get a working code example plus its source, directly from the CLI.

**This is a port, not a copy — it does not have to be 1:1.** As part of bringing it over:

- **Slot elegantly into `winapp`** — follow existing CLI patterns (command/option shape, help text, `--json` output, error handling, exit codes) so it feels native, not bolted on.
- **Make it more efficient where possible** — simplify/trim the ported logic; reuse existing `winapp` helpers/services instead of duplicating them.
- **Keep it light** — mind payload and startup cost; keep the command lean/AOT-friendly, consistent with `winapp`.
- **Keep the UX sensible** — the command, flags, and output should make sense within `winapp`'s conventions, not just a lift-and-shift of the old CLI's surface.

Scope matches what `winui-search` covers today (WinUI Gallery + Windows Community Toolkit).

## Out of scope

- Semantic / natural-language search over the corpus — tracked separately in **#625**, which builds on this port.

## Open questions

- [x] Final command/flag surface within `winapp` (align with `winapp` conventions).
- [x] Which parts of `winui-search` to reuse vs. re-implement against existing `winapp` services.
- [ ] Result / `--json` schema.

## Additional context

Foundation for **#625** (semantic `winapp find-ui`), which depends on this issue.

Once the capability lands in `winapp`, **remove the old `winui-search` references in microsoft/win-dev-skills** (`src/tools/winui-search/` and any skills/docs that point at it) so there is **no duplicated code** across the two repos and `winapp` is the single source of truth.

Jaylyn-Barbee:
## Description

Adds **`winapp find-ui`** — a WinUI-only search over WinUI controls and samples that returns a working code example (XAML + C#) plus its source. It helps an gent (or a human) find the right control and its real usage instead of guessing an API.

The corpus spans three sources: the **WinUI 3 Gallery**, the **Windows Community Toolkit**, and the **microsoft-ui-reactor ReactorGallery** (C#-only declarative WinUI), plus a few curated core patterns. Workflow is search → fetch full code with `--id`. `--json` gives structured output for agents.

## Usage Example

```bash
# Find a control, then fetch its full XAML + C#
winapp find-ui "tabbed layout"
winapp find-ui --id gallery-tabview-1

# Batch across sources; restrict to one source; structured output
winapp find-ui --id gallery-tabview-1 --id toolkit-settingscard-1 --id reactor-flex-1
winapp find-ui "flex layout" --source reactor
winapp find-ui "color picker" --json
```

## Related Issue
Closes https://github.com/microsoft/winappCli/issues/651

## Type of Change

- ✨ New feature

## Checklist

- [x] Main [README.md](../README.md) updated (if applicable)
- [ ] [Language-specific guides](../docs/guides) updated (if applicable) — n/a
- [ ] [Sample projects updated](../samples) to reflect changes (if applicable) — n/a
- [x] Agent skill templates updated in `docs/fragments/skills/` (if CLI commands/workflows changed)

## Additional Notes

- **No embedded corpus by design.** find-ui fetches the Gallery/Toolkit/Reactor data from GitHub on first use and caches it per-user under `<global .winapp>/cache/find-ui/{gallery,toolkit,reactor}` (7-day TTL, or `--refresh`). Only small curated tag/keyword enrichment ships in the binary, so the **first run requires network access**. Each source is an `ISearchProvider` behind a shared `CachedProviderBase`; adding a source is one provider + one registry line.
- **Testing:** full C# suite **3,638 passed / 4 skipped**; the 1 failure (`AnalyzeDumpAsync_ManagedDumpWithDeepStack_...`) is a pre-existing ARM64-host-analyzing-an-x64-dump environmental failure, unrelated to this change. Branch was rebased onto current `main` (no conflicts); docs validated in sync.
- New  find-ui  usage telemetry ( FindUiUsageEvent ) — please flag for telemetry/privacy review: emits only bounded, non-PII values (mode, registry-validated  --source , corpus-canonical resolved ids, counts; the free-form query is never captured and unresolved ids are counted only), at  LogLevel.Measure  pending Critical-event approval, and is a deliberate deviation from the centralized-telemetry convention because  --id  can only be validated against the loaded corpus in the handler.


nmetulev:

## PR Review — `jay/find-ui-port` vs `main` (26 commits, 51 files, +8036/-4)

**Summary:** Critical 0 · High 3 · Medium 3 · Low 0 (gut check kept 6 of 19)

| ID | Location | Domain | Issue |
|---|---|---|---|
| H1 | `Services/Controls/GalleryFetcher.cs:477-599`, `ToolkitFetcher.cs:938-996` | correctness | `--id` emits XAML that does not parse and C# that does not compile |
| H2 | `Commands/FindUiCommand.cs:293-321` | security | Terminal escape sequences from fetched content reach raw stdout |
| H3 | `docs/npm-usage.md` | ship-surfaces | Generated doc not regenerated; `findUi()` undocumented |
| M1 | `Commands/FindUiCommand.cs:59-181` | cli-ux | `--max` silently ignored with `--list` / `--id` |
| M2 | `docs/usage.md:940` | ship-surfaces | `--json` shape documented wrong |
| M3 | `Services/Controls/SearchProvider.cs:260-308` | necessity | Reactor opt-in source adds permanent surface |

**Must fix before merge:** H1, H2, H3
**Safe to decline:** M1 (fine as documented behavior) · M2 (one-line doc fix) · M3 (product call, not a defect)

---

### H1 · `--id` emits code that does not parse — the feature core output
**high · confidence high · validated**

The value proposition is "give an agent a working code example." It does not, on common controls, from a warm cache:

```
$ winapp find-ui --id gallery-appbarbutton-5
  <AppBarButton.KeyboardAccelerators>
    <KeyboardAccelerator Modifiers="Control" Key="S"/>
  <AppBarButton.KeyboardAccelerators/>          <-- fabricated self-closing tag
</AppBarButton></AppBarButton.KeyboardAccelerators></AppBarButton>

XmlException: 'AppBarButton.KeyboardAccelerators' start tag ... does not match
the end tag of 'AppBarButton'.
```

This sample is short and **not truncated** — the tag-balancing "repair" (`GalleryFetcher.cs:519-534`) corrupts already well-formed input.

```
$ winapp find-ui --id gallery-storagepickers-5
      ThumbnailDetailsTextBlock.Text =
}}}                                            <-- statement cut mid-assignment
```

Contributing causes:
- `TruncateXaml` boundary check is a tautology (`lastGt` is always the `>` just found, so `if (lastLt < lastGt) break;` always fires), so it cuts inside quoted attributes/comments.
- `ToolkitFetcher.TruncateCSharp` has no brace balancing at all, unlike `GalleryFetcher.TruncateCode`.
- `TruncateXaml`/`AnyTagRegex` are duplicated across the two fetchers and have already drifted (only the Gallery copy strips comments).

**Recommendation** — validate before emitting: run the cleaned XAML through `XmlReader`; if it does not parse, omit the block with a "see full sample" note rather than shipping broken markup. Same for C#: cut only at a complete syntactic boundary, otherwise omit. Extract the shared truncation into one file so the copies cannot drift again. Note: simply reusing `GalleryFetcher.TruncateCode` in Toolkit is not sufficient — it still emits incomplete statements.

### H2 · Terminal escape sequences from fetched content reach raw stdout
**high · confidence high · validated**

`console.WriteLine(line)` (line 318) writes code/free text literally. `Markup.Escape` only neutralizes Spectre `[`/`]`, not ESC/OSC bytes, and nothing in the fetch → clean → truncate pipeline filters control characters.

Red-teamed on the built binary with a poisoned corpus entry (`ESC]0;PWNED-BY-UPSTREAM BEL ESC[41m...`):

```
stdout bytes=4172  ESC=3  BEL=1  PWNED=True
```

The OSC title-set sequence survives intact — in a real terminal it rewrites the window title. Reachable if a booby-trapped sample lands in WinUI-Gallery / CommunityToolkit `main`; this output is explicitly designed to be piped into agents and terminals.

**Recommendation** — strip C0/C1 control characters (below `0x20` except `\n`/`\t`) once at the `Scenario` construction boundary, so console, cache, and JSON paths are all covered by one call.

### H3 · `docs/npm-usage.md` is stale — `findUi()` undocumented
**high · confidence high · validated**

```
$ node scripts/generate-docs.mjs --check
[generate-docs] docs/npm-usage.md is out of date.
EXIT=1
```

`winapp-commands.ts` exports `findUi()`; `docs/npm-usage.md` has zero matches for `findUi`/`find-ui`. `npm run build` runs `generate-commands` but not `generate-docs`, and no workflow runs `generate-docs:check` — so this drifts silently.

**Recommendation** — run `cd src/winapp-npm && npm run generate-docs` and commit. Optionally add `generate-docs:check` next to the existing `generate-commands` step in `build`.

### M1 · `--max` silently ignored with `--list` / `--id`
**medium · validated**

```
--list lines=495 ; --list --max 2 lines=495   -> --max ignored: True
$ winapp find-ui --list --source gallery
❌ --source only applies to search; it cannot be combined with --list or --id.
```

`--source` is rejected in that position "so scripted callers do not get a false sense of filtering" — `--max` gets a free pass under the same rationale. Either add it to that check, or scope its description to search. Declinable if you would rather keep the flag permissive.

### M2 · `--json` shape documented wrong
**medium · validated**

`docs/usage.md:940` says matches carry `` (`id`, `control`, `source`, `header`, `score`) ``. Actual:

```
match keys:    source, control, score, description, scenarios
scenario keys: id, header
```

`id`/`header` are nested per-scenario. `docs/fragments/skills/winapp-cli/find-ui.md:76` already has it right — copy that wording.

### M3 · Reactor opt-in source
**medium · static-only · disputed by cross-check**

Added late from review feedback; costs `--source reactor`, `reactor-*` ids, dual memoized engines in `ControlsSearchService`, and `includeReactor` threaded through the command and telemetry — for an incubation project with plausibly near-zero users. The necessity reviewer recommended cutting it and re-adding on request; the cross-check disputed that as product-scope preference rather than a defect. Maintainer call; not a merge blocker.

---

### Coverage notes

- **Gut check dropped 13 of 19.** Cut as true-but-not-necessary: the `sampleDef` `..` URL-join (bounded to the same public host), `--source` as an enum instead of `string`, three inconsistent `User-Agent` strings, `ToolkitFetcher.cs` at ~1100 lines and `GalleryFetcher.cs`/`SearchEngine.cs` past the 800-line soft limit (real rule, zero user impact — worth doing on the next touch), the bespoke `FindUiUsageEvent` (documented as a deliberate reviewer-flagged deviation), corrupt-`keywords.json` degradation, and the `--id` colorizer / fetch-notice polish.
- **Verified clean:** `cli-schema.json` ↔ command definition, `winapp-commands.ts` schema version, `.claude/` ↔ `.github/plugin/` mirror, `$SkillDescriptions` entry, `docs/telemetry.md` ↔ `FindUiUsageEvent`, `README.md`, agent decision tree + command reference, `version.json` (correctly untouched), embedded `Data/*.json` via `LogicalName` (AOT-safe), and the scoped `.editorconfig` (style/perf only). No new `PackageReference`s. `find-ui` correctly falls through the npm `cli.ts` generic passthrough.
- **Build:** `dotnet build -c Debug` — 0 warnings, 0 errors. Cross-check ran 103 .NET + 223 npm tests, all passing.
- **Review method:** 6 parallel specialist reviewers (security, correctness/tests, cli-ux, alternative-solution, ship-surfaces, necessity) plus a different-model-family cross-check, then runtime validation against the built binary.

nmetulev:

## Addendum — corrected H1 framing, plus what to do now vs. after merge

Follow-up to my review above after digging into the upstream data sources. Two corrections and a concrete plan.

### Correction 1: H1 is inherited, not a regression

I implied this PR broke the XAML output. It didn't. I validated both corpora with an XML parser (declaring all prefixes so undeclared-prefix isn't a false positive):

```
win-dev-skills winui-search baked snapshot : 321 scenarios, 9 structurally malformed
this PR (fetched corpus)                   : 321 scenarios, 9 structurally malformed
same ids: xamlresources-3, appbarbutton-5, swipecontrol-4, swipecontrol-5,
          listview-8, infobadge-2, selectorbar-3, themeshadow-1, richtextblock-3
```

Byte-identical outcome — ~3% of Gallery scenarios. **H1 is a pre-existing bug in the ported code, not something this PR introduced.** It's still worth fixing (a snippet that doesn't parse is worse than no snippet, for an agent-facing tool), but it should not be read as a regression, and I'm sorry for the misleading framing.

Worth noting *why* it went unnoticed: upstream bakes its corpus into a checked-in `Data/gallery-scenarios.json`, so the broken output is at least a file someone could inspect. This PR regenerates it on each user's machine, where nobody ever sees it. Same 9 bugs, less visible.

### Correction 2: dropping the baked snapshot didn't make things simpler

Upstream ships auto-baked `Data/*-scenarios.json` (~737 KB across gallery/toolkit/reactor) plus a `BackgroundUpdater`; this port drops all of it in favour of always-fetch. That legitimately removes ~737 KB from the repo — but the cost landed in code:

| `SearchProvider.cs` | upstream | this PR |
|---|---|---|
| code lines | 113 | **173** (+53%) |
| comment lines | 46 | **95** (+107%) |

Upstream's whole load path is 11 lines (`cache → embedded → done`, never blocks on network). Here it's ~50, because once the guaranteed fallback is gone every caller has to handle "no data": try/catch→Empty, a TTL-ignoring stale re-read, `ControlsDataUnavailableException`, `allowCoreOnly` so `--list` works offline, the one-shot "fetching…" notice, and dual-engine memoization. Not a blocker for this PR — just worth knowing that the simplification went the other way.

---

## Do now (in this PR, small)

**1. Validate at the corpus boundary — one function, closes H1 and H2 together.** At `Scenario` construction: `XmlReader`-parse the XAML, strip C0/C1 control characters, check C# brace balance; drop or flag what fails. ~30 lines, no upstream dependency. Today those 9 broken scenarios ship silently and the ANSI-injection path (H2) is open.

**2. Use `[ToolkitSample(id:, title:, description:)]` instead of stripping it.** `ToolkitFetcher.CleanCSharp` regexes the attribute away as noise, then re-derives the same mapping from a 34-entry hardcoded `SampleOverrides` table plus naming heuristics. Upstream already declares it in the file we download:

```csharp
[ToolkitSample(id: nameof(SettingsCardSample), "SettingsCard",
    description: "A sample for showing how SettingsCard can be static or clickable.")]
```

vs. what we hand-maintain:

```csharp
["SettingsCardSample"] = ("settingscard", "SettingsCard"),
```

Parsing it before stripping deletes the override table and the heuristics, and stops the drift where a newly added Toolkit sample gets a guessed name until someone notices.

**3. Dedupe the two `TruncateXaml` / `AnyTagRegex` copies** (already covered in H1 — noting it here since it's the same edit).

H3 (regenerate `docs/npm-usage.md`) still stands as-is.

## Do after this merges

Filed #703 for the follow-up work.

Short version: the Gallery/Toolkit fetchers are ~1,450 lines of parsing that exists only to reconstruct data those repos already have. Reactor's fetcher is 7 KB for the same job because the Reactor team publishes `reactor-search-index.json`. The fix is to get the other two to publish an index too — and the Toolkit **already generates exactly that metadata** at build time via `ToolkitSampleMetadataGenerator`, it just never leaves the compiler. Once an index exists, those fetchers collapse to ~100 lines and this class of malformed-output bug disappears by construction.

Also worth considering separately (not required for merge): baking the corpus in winappCli CI and embedding it. It adds bytes rather than removing them, but Brotli on the real corpus measures **575 KB → 57 KB** (gallery) and **106 KB → 12 KB** (toolkit) — ~70 KB total — and it turns item 1 above into a **CI gate** instead of a silent runtime event, so an upstream reshuffle breaks our build with a reviewable diff instead of quietly degrading users' output.

nmetulev:

## H4 (new, high) — a source that fails to load is reported as "no match"

Found while investigating why `--source reactor` returned nothing. This is **not** Reactor-specific and is independent of any one source being reachable.

### Repro

Warm `gallery` + `toolkit` cache, no `reactor` cache (i.e. the Reactor fetch failed at some point):

```
$ winapp find-ui "flex layout"                    → returns toolkit results     exit 0
$ winapp find-ui "flex layout" --source reactor   → No WinUI controls matched   exit 1
$ winapp find-ui --id reactor-flex-1              → Pattern not found           exit 1
```

Identical with `gallery` removed instead and `toolkit` cached:

```
$ winapp find-ui "button" --source gallery        → No WinUI controls matched   exit 1
$ winapp find-ui --id gallery-button-1            → Pattern not found           exit 1
```

The corpus is demonstrably loaded and working — the first command succeeds in the same state. Only the *filtered* query is wrong.

### Root cause — `Services/Controls/ControlsSearchService.cs:129-159`

```csharp
if (data.Scenarios.Length == 0)
{
    anyProviderEmpty = true;   // captured...
    continue;
}
...
if (allScenarios.Count == 0)   // ...but this only fires when EVERY provider is empty
{
    throw new ControlsDataUnavailableException("...connect to the internet...");
}
```

`anyProviderEmpty` is only read at line 172, to decide whether to memoize; it never reaches the caller. So the friendly "connect and run once" error requires **all** providers to be empty. One warm cache masks every other source's failure, and `EmitSearch` then filters `sourceFilter: "reactor"` across a corpus containing no reactor scenarios and truthfully reports zero hits.

`--refresh` does not help — the fetch-failure path also returns `ProviderData.Empty`.

### Why this is high

`find-ui` is explicitly aimed at agents. An agent runs `--source reactor` for "flex layout", receives a confident *no match*, and concludes Reactor has no flex layout — when Reactor ships 93 controls, `flex` among them. A silent wrong answer is worse than an error here, because nothing downstream can detect it.

Triggers for real users: offline, behind a proxy that blocks `raw.githubusercontent.com`, during a GitHub outage, or after an upstream path rename. Note the last one is not hypothetical — `CacheVersion` `"17"` records exactly such a Gallery move.

### Smallest fix — ~5 lines, no signature changes

`SearchEngine` already holds the scenario array:

```csharp
internal bool HasSource(string source) => _scenarios.Any(s => s.Source == source);
```

Then in `EmitSearch` / `EmitCode`, before reporting a miss: if the requested `--source` (or the `--id` prefix's source) has no scenarios loaded, `Fail(...)` with the existing "connect and run once" message, naming that source. No new service plumbing or extra state required.

A test worth adding alongside it: fake providers where one returns `Empty` and another returns data, asserting `--source <empty>` fails rather than returning zero matches. `FakeControlsSearchService` already supports this shape.

---

Updated summary for the review as a whole: **Critical 0 · High 4 · Medium 3**. Must fix before merge: H1, H2, H3, H4.

nmetulev:

## Network profiling: find-ui is non-functional in agent sandboxes

Filed as #704, summarised here because it changes the merge calculus.

`find-ui` fetches everything from `raw.githubusercontent.com`. Measured from a sandboxed agent environment:

```
raw.githubusercontent.com        FAIL (20s timeout)
github.com/.../raw/...           FAIL (redirects to raw.githubusercontent.com)
api.github.com  (contents+tree)  HTTP 200, ~200ms
codeload.github.com              HTTP 200
```

Standard allowlist pattern — `api.github.com` permitted, `raw.githubusercontent.com` not. Agent sandboxes commonly look exactly like this, and this command is aimed at agents.

| Fetcher | Index | Content | Works? |
|---|---|---|---|
| Gallery | `raw` ❌ | `raw` ❌ | no |
| Reactor | `raw` ❌ | — | no |
| Toolkit | `api.github.com` tree ✅ | `raw` ❌ | no — dies after the tree |

**In fairness, the fully-cold-cache path fails loudly and correctly:**

```
❌ No WinUI control data is available. find-ui fetches ... connect to the
   internet and run the command once to populate the local cache.
exit 1
```

That's honest behaviour, so on its own I'd call this a known limitation rather than a merge blocker. **The problem is the interaction with H4** — with a partially warm cache, the same condition is reported as a confident "no match" instead. Together they produce silently wrong answers to an agent, which is the one failure mode this command cannot afford.

Switching hosts isn't the fix: `api.github.com` is rate-limited to 60 req/hr unauthenticated, and a cold fetch measures ~540 requests (Gallery up to 433, Toolkit ~105, Reactor 1).

### Recommendation

**H4 must be fixed before merge** — unchanged from my previous comment.

**For the proxy case, I'd suggest baking and embedding the corpus before merge as well** (#704), not because loud failure is unacceptable, but because that one change resolves four things at once: proxy/offline operation, H1's invisibility (validating snippets becomes a CI gate with a reviewable diff instead of a silent per-user event), the cache-fallback complexity (`SearchProvider.cs` drops back toward upstream's 113 lines from 173), and cold-start latency. It measures ~70 KB embedded with Brotli (gallery 575 KB → 57 KB, toolkit 106 KB → 12 KB).

Doing it now is also less total work than retrofitting once users have caches in the wild — and the original `winui-search` already shipped exactly this design, so it's a restore rather than a new build.

Worth keeping the framing **embedded snapshot as a floor, network refresh as the ceiling** (upstream's embedded + `BackgroundUpdater` model), so this doesn't trade a proxy bug for a staleness bug.

That's a maintainer call on scope, not a defect — happy to be overruled on sequencing. H4 I'd hold firm on.

nmetulev:

@Jaylyn-Barbee — re-reviewed at `9cbd0e79`. **Everything from my earlier passes is fixed.** One last bug and this is good to merge from my side.

## H5 (new, high) — plain-text code blocks are word-wrapped to console width, producing invalid XAML

`WriteScenarioContent` writes code lines through `console.WriteLine(line)` (`Commands/FindUiCommand.cs:348`). Spectre word-wraps that to the console width, so the emitted snippet is **re-flowed**, not verbatim.

### Repro

```powershell
winapp find-ui --id gallery-gridview-2 > out.txt
```

Redirecting means no TTY, so Spectre falls back to 80 columns and the wrap lands mid-tag:

```
SystemControlPageTextBaseMediumBrush}"/></StackPanel></StackPanel></Grid></DataT
emplate>
```

`</DataTemplate>` split into `</DataT` + `emplate>` — the XAML no longer parses.

### Measured

| Path | Lines | Max line length |
|---|---|---|
| `--json` (canonical) | 39 | 127 |
| plain text | 47 | 80 |

**8 line breaks introduced purely by rendering**, and every line capped at the console width.

This is easy to miss on a wide terminal: at 120 columns the same snippet breaks at a space, so it still looks correct. It only corrupts when a break point lands inside a token — which is exactly what happens at 80.

### Why it matters

It fires whenever there's no wide TTY: piping to a file, CI logs, or any terminal under ~130 columns. Since it's width-dependent it will reproduce intermittently and look like a corpus bug rather than a rendering bug.

`--json` is unaffected, so agents on the JSON path are safe — but the plain path is what a human copies out of the terminal.

### Suggested fix

Bypass markup/wrapping for fenced-block content — e.g. `console.Write(new Text(line))`, or write those lines straight to `Console.Out`. The code path at line 348 is already separate from the `MarkupLine` paths above it, so this is a narrow change.

Worth a regression test asserting that the emitted line count for a known scenario matches the canonical (`--json`) content, which would have caught this independently of terminal width.

---

## Everything else verified fixed

| Finding | Status |
|---|---|
| **H1** malformed XAML | ✅ 1 of 301 remains (`gallery-customxamlconditionals-3`, uses `x:Key` with no `xmlns:x` — inherited from the upstream sample) |
| **H2** ANSI/control-char injection | ✅ `ScenarioSanitizer` at the corpus boundary |
| **H3** stale `docs/npm-usage.md` | ✅ `findUi()` documented; regenerating produces no diff (`--check` fails on line endings only) |
| **H4** unloaded source reported as "no match" | ✅ now errors: `❌ No 'reactor' control data is available locally…` exit 1 |
| **M2** wrong `--json` shape in docs | ✅ corrected |
| Truncation dedupe | ✅ shared `ControlSnippetText.cs` |

Also verified `--source core` now skips the network entirely (0.2s behind a black-hole proxy) — nice addition.

**Tests:** 227 find-ui-related tests, 0 failures. Build clean, 0 warnings.

**Still open, and I'd decline both:**
- **M1** — `--max` silently ignored with `--list` (495 lines either way). Fine as documented behavior.
- The proxy/sandbox limitation (#704) and upstream indexes (#703) are tracked separately and are not merge blockers.

**Verdict: fix H5 and this is good to merge.**

nmetulev (approved):

PR #721, opened 2026-08-07
Add winapp-pr -Uninstall and bound the download cache

nmetulev:
Stacked on #720 — merge that first; this branch contains its commit.

There was no way back to a clean state. Installing a build leaves a package, a trusted CI certificate, cached downloads, and install records, and only the certificates had a cleanup command. Nothing ever removed the cache — on my machine it had reached **771 MB across 30 run folders**.

## `-Uninstall`

```powershell
winapp-pr -Uninstall
```

```
>> Uninstalling
   This will remove:
     - the installed package (winapp-dev 0.5.1.64)
     - 1 CI certificate(s) trusted for it
     - 771 MB of cached downloads
   Continue? [Y/n]
```

It lists what it will remove and asks first; `-Force` skips the prompt. `winapp-pr` itself stays, so you can go straight to another build. `-Uninstall -All` also removes the tool and its PATH entry, for putting a machine back exactly as it was.

Certificate removal moves into a helper shared with `-PruneCerts`, so both paths elevate once and neither touches certificates this tool did not add. Uninstall removes only the tracked thumbprints — unrelated `CN=runneradmin` certificates from other tools are left alone, which matters because `TrustedPeople` is a shared store.

`-All` deletes the directory the running script lives in. That works because PowerShell reads a `.ps1` into memory rather than locking it, and the result is verified rather than assumed — if the directory survives, it says so instead of reporting success.

## Bounded cache

Deleting the download after installing would be the obvious fix for the 771 MB, but the cache does buy something measurable: re-installing a build costs about 7 seconds of download, and flipping between two PRs is a normal thing to do while reviewing.

So the few most recent builds are retained and the rest dropped after each install, with an entry refreshed when it is reused so retention follows the builds you actually switch between. Steady state is around 48 MB rather than unbounded:

```
   [OK] Downloaded winappcli-jay-find-ui-port_0.5.1.64_arm64.msix
   Pruned 2 older cached build(s), freeing 32 MB.
```

## Validation

Run end to end on a real install. `-Uninstall` removed the package, untrusted its certificate, and cleared the cache and records, leaving the 3 untracked CI certificates in place and `winapp-pr` still working; `-Status` then correctly reported no package installed. `-All` was run *from the installed copy* to confirm it can delete itself, after which the install directory and PATH entry were both gone. The original build was reinstalled afterwards and verified.

Cancelling at the prompt makes no changes, and the nothing-to-remove case reports that rather than printing an empty list.

Cache retention was checked as a unit (six entries reduced to the current one plus the two most recent) and end to end, seeding older entries alongside a real one and confirming an install pruned 80 MB down to 48 MB while keeping the build just installed.

### DIFF
--- a/scripts/winapp-pr.ps1
+++ b/scripts/winapp-pr.ps1
@@ -80,6 +80,9 @@ $GhExe = 'gh'
 
 $ToolHome     = Join-Path $env:LOCALAPPDATA 'winapp-dev'
 $CacheRoot    = Join-Path $ToolHome 'cache'
+# Enough to flip between a couple of builds without re-downloading, while staying bounded.
+# Each run folder holds one architecture, around 16 MB.
+$CacheKeep    = 3
 $StateFile    = Join-Path $ToolHome 'current.json'
 $TrustedFile  = Join-Path $ToolHome 'trusted-certs.json'
 $InstallDir   = Join-Path $env:LOCALAPPDATA 'Programs\winapp-dev'
@@ -681,6 +684,9 @@ function Get-CachedMsixPackage {
     }
     if ($cached) {
         Write-Ok "Using cached package: $($cached.Name)"
+        # Mark as most recently used so retention favours the builds you actually switch between.
+        (Get-Item $runCache).LastWriteTime = Get-Date
+        Remove-OldCacheEntries -KeepPath $runCache
         return $cached
     }
 
@@ -705,6 +711,7 @@ function Get-CachedMsixPackage {
 
     Get-ChildItem $runCache -File | ForEach-Object { Unblock-File $_.FullName -ErrorAction SilentlyContinue }
     Write-Ok "Downloaded $($package.Name)"
+    Remove-OldCacheEntries -KeepPath $runCache
     return $package
 }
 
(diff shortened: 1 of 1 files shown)


zateutsch:

## PR Review — `nmetulev-winapp-pr-uninstall` vs `origin/main`

**Summary:** 0 critical, 3 high, 2 medium, 0 low. Gut check kept 5 of 6 candidates.

### Recommended action

**Must fix before merge:** H1, H2, H3  
**Safe to decline:** M1, M2 — stale metadata is recoverable; M2 does not close the pre-existing writable-script boundary.

### Findings

#### H1 — `scripts/winapp-pr.ps1:1017-1056` — `-All` claims full removal while trusted certs remain

- **Severity:** high
- **Confidence:** high
- **Validation:** static-only; reproducing requires a tracked machine certificate and declining/failing UAC.
- **Domain:** cli-ux
- **Multi-model:** confirmed
- **Finding:** If certificate removal fails, `-Uninstall -All` deletes the tool needed to retry cleanup and still prints “fully removed.”
- **Evidence:** The failure path preserves `$stillTrusted` and tells the user to run `winapp-pr -PruneCerts later`, but execution then removes `$InstallDir` and reports full removal.
- **Recommendation:** When `$stillTrusted` is non-empty, preserve the tool and tell the user to rerun cleanup.

#### H2 — `scripts/winapp-pr.ps1:1202-1205` — destructive public command has no tests

- **Severity:** high
- **Confidence:** high
- **Validation:** validated; the new public help surface runs, and repository test search found no coverage for `winapp-pr`, `Invoke-Uninstall`, or its cleanup helpers.
- **Domain:** correctness
- **Multi-model:** severity disputed (medium)
- **Finding:** The destructive uninstall command has no automated coverage for cancellation, forced cleanup, partial certificate failure, or `-All` self-removal.
- **Recommendation:** Add focused Pester tests with mocked package, certificate, PATH, and filesystem operations.

#### H3 — `scripts/winapp-pr.ps1:842-871,917-945` — failed retirement can permanently lose certificate tracking

- **Severity:** high
- **Confidence:** medium
- **Validation:** static-only; confirmation requires a machine certificate whose retirement fails while the new import succeeds.
- **Domain:** security
- **Multi-model:** independently found
- **Finding:** Certificate retirement suppresses removal errors, then overwrites tracking with only the new thumbprint; the changed prune logic can no longer discover the leftover trust anchor.
- **Evidence:** Line 845 uses `-ErrorAction SilentlyContinue`, line 871 replaces the tracked set, and lines 922-925 now prune tracked thumbprints only.
- **Recommendation:** After elevation, query the store and retain any retirement targets still present in `trusted-certs.json`.

#### M1 — `scripts/winapp-pr.ps1:994-996` — stale install state survives uninstall

- **Severity:** medium
- **Confidence:** high
- **Validation:** static-only
- **Domain:** correctness
- **Finding:** If only `current.json` remains, uninstall reports “Nothing to remove” and leaves stale source metadata that a later `-Update` can reuse.
- **Recommendation:** Include `$StateFile`/`$ToolHome` existence in the early-return condition.

#### M2 — `scripts/winapp-pr.ps1:901-905` — elevated shell is resolved through PATH

- **Severity:** medium
- **Confidence:** high
- **Validation:** static-only; red-teaming would require approving elevation of a harmless PATH-shadowed executable.
- **Domain:** security
- **Finding:** The new cleanup path elevates a bare `pwsh`/`powershell` name resolved through user-controlled PATH.
- **Recommendation:** Use the trusted absolute current-host executable path at both elevation sites.

### Coverage

- security: 2 findings
- correctness: 2 findings
- cli-ux: 1 finding
- alternative-solution: clean
- ship-surfaces: clean
- necessity/simplicity: scope justified
- multi-model: H1 confirmed; H3 independently found (Opus)
- validation: script parsing and `-Help` succeeded. The full build was blocked by the local NativeAOT toolchain failing to resolve `vswhere.exe`/the linker; the worktree remained clean.

The default-yes confirmation candidate was dropped in the gut check because the user explicitly invoked uninstall, sees the removal list, and can use `-Force` for automation.

zateutsch (commented):
Attached some review findings above. Not a ton but some high severity findings, H1 definitely seems like it could be an issue.

PR #724, opened 2026-08-10
Make dotnet run pass its arguments to the app, as it does without the package

azchohfi:
> **Stacked on #731** — merge that first. This PR makes MSBuild properties the only way to configure winapp through `dotnet run`, and #731 is what supplies `Detach`, `UnregisterOnExit`, `Clean`, `Symbols`, `Executable`, and `RunArgs`.

## Problem

For a project that references this package, both of these fail:

```
> dotnet run --devtools
❌ Unrecognized argument: '--devtools'. To pass arguments to the app, use --

> dotnet run -- --devtools
❌ Unrecognized argument: '--devtools'. To pass arguments to the app, use --
```

The second one is the trap: the error tells you to use `--`, and you *did*. Both commands work on any project that doesn't reference the package, so adding it silently changed what `dotnet run` means.

## Why both spellings fail identically

The .NET SDK appends your application arguments to `$(RunArguments)` verbatim, and System.CommandLine consumes any standalone `--` while parsing `dotnet run` itself and never re-emits it ([`RunProperties.WithApplicationArguments`](https://github.com/dotnet/sdk/blob/main/src/Cli/dotnet/Commands/Run/RunProperties.cs) → [`CommonRunHelpers.CombineRunArguments`](https://github.com/dotnet/sdk/blob/main/src/Cli/dotnet/Commands/Run/CommonRunHelpers.cs)).

Verified with a probe app that echoes its argv — `dotnet run --devtools` and `dotnet run -- --devtools` produce **byte-identical** results. The separator is unrecoverable, so the targets file cannot tell the two apart.

Because the package overrides `RunCommand` to an intermediary launcher that has *its own options*, those tokens land in winapp's option namespace instead of your app's.

## Fix

One line. `RunArguments` now ends with a separator:

```xml
<RunArguments>$(_WinAppRunArgs) --</RunArguments>
```

Everything the SDK appends lands in winapp's passthrough region and reaches your app unchanged.

**No CLI change is needed** — `winapp run` keeps rejecting unknown options, so a typo like `winapp run . --debug-outpt` still fails loudly in a hand-typed invocation.

| Command | Before | After |
|---|---|---|
| `dotnet run --devtools` | ❌ error | app gets `--devtools` ✅ |
| `dotnet run -- --devtools` | ❌ error | app gets `--devtools` ✅ |
| `dotnet run --detach` | winapp detaches | **app gets `--detach`** ⚠️ |
| `dotnet run -p:WinAppRunDetach=true` | winapp detaches | unchanged ✅ |
| `winapp run . --detach` (direct CLI) | winapp detaches | unchanged ✅ |
| `winapp run . --debug-outpt` (direct CLI) | ❌ typo caught | ❌ typo caught ✅ |

## ⚠️ Breaking change

Options written after `dotnet run` used to configure the launcher. `dotnet run --detach` now passes `--detach` to your app instead. Use the properties: `-p:WinAppRunDetach=true`.

MSBuild can't warn about this, because it never sees those tokens — so winapp does. When invoked as the NuGet caller and a forwarded argument matches one of its own options:

```
ℹ '--detach' was passed to your application, not to winapp.
  To configure winapp, use -p:WinAppRunDetach=true instead.
```

Deliberately narrow: unknown arguments stay silent (they never had a winapp meaning, so a notice would be noise on every run), a direct `winapp run . -- --detach` stays silent (an explicit request to forward), and attached-value spellings such as `--executable=foo.exe` are matched on the option name, since those configured winapp before this change too.

Note that a forwarded `--json` or `--quiet` **does** produce the notice: with the separator in place those tokens go to your app rather than putting winapp into JSON/quiet mode, so the notice is accurate. The suppression guard still applies when winapp itself is in that mode (`winapp run --json`, or `-p:WinAppRunArgs="--json"`).

## Validation

- 132/132 `RunCommandTests` pass, including 4 new tests covering each notice case.
- 27/29 NuGet tests pass; the 2 failures are the pre-existing package-layout tests needing a built `.nupkg`.
- New test asserts the shared `_WinAppRunArgs` stays separator-free, so `RunPackagedApp` is provably unaffected.
- Verified on the built binary: notice fires for a known option, stays silent for `--devtools` and for direct CLI use, and covers attached-value forms like `--executable=foo.exe`.

Docs updated across `docs/dotnet-run-support.md`, `docs/usage.md`, `docs/guides/dotnet.md`, the package README, and the frameworks skill.


azchohfi:
@copilot review

azchohfi:
Addressing the three suppressed comments from the last Copilot review, since they don't create reply-able threads. All three were legitimate; two were fixed and one was a documentation error on my side rather than a behavior bug.

**1. Attached-value spellings missed the notice** — *fixed in `c0e4de79`.*

Confirmed on the built binary: `--executable=foo.exe` produced no notice at all, so an invocation that previously configured winapp changed meaning silently. Lookup, de-duplication, and property mapping now use the name before the `=`, while the message still quotes the token as typed:

```
ℹ '--executable=foo.exe' was passed to your application, not to winapp.
  To configure winapp, use -p:WinAppRunExecutable=<path> instead.
ℹ '--detach=true' was passed to your application, not to winapp.
  To configure winapp, use -p:WinAppRunDetach=true instead.
```

`--devtools=1` still stays silent, so unknown options don't start generating noise. Covered by a new test.

**2. Docs said "in a previous release"** — *fixed in `c0e4de79`.* That was plainly wrong, since this PR introduces the change; it now reads as a breaking change in this release and is marked `[!IMPORTANT]`.

**3. `--json` / `--quiet` suppression claim** — *PR description corrected; no behavior change.*

You were right that the claim was false, and I verified it:

```
> winapp run . --caller nuget-package -- --json
ℹ '--json' was passed to your application, not to winapp. ...
```

But the behavior is correct as-is. With the separator in place, `dotnet run --json` sends `--json` to the *app* rather than putting winapp in JSON mode, so `isJson` is legitimately false and the notice is accurate — that forwarded token really did go to the application. There's also no JSON document to corrupt, because winapp isn't emitting one. The guard still does its job when winapp itself is in JSON or quiet mode (`winapp run --json`, or `-p:WinAppRunArgs="--json"`). So only my PR description was wrong, and it's been updated.

Two later commits also landed here: `5fd410f8` and `b33bcb16` cherry-pick #737, which fixes a flaky PID-file test I introduced in #732 that was failing `build-and-package` on this branch and on `main`.

azchohfi:
@copilot review

docs/dotnet-run-support.md
```diff
@@ -182,6 +181,42 @@ The main build script now includes NuGet packaging:
...
+`dotnet run` behaves the same way whether or not a project references this package: **everything you
+write after `dotnet run` is passed to your application.**
+
+```powershell
+dotnet run --devtools          # your app receives --devtools
+dotnet run -- --devtools       # identical
+dotnet run --detach            # your app receives --detach
+```
+
+A standalone `--` is optional and has no effect. The .NET SDK consumes it while parsing its own
```
nmetulev: `--` is not optional when an application argument collides with a `dotnet run` option. With the real SDK, `dotnet run --configuration Release` is consumed by the SDK; only `dotnet run -- --configuration Release` forwards those tokens. Please qualify this section and the equivalent claims in `docs/guides/dotnet.md`, `docs/usage.md`, `src/winapp-NuGet/README.md`, and the framework skill.

docs/guides/dotnet.md
```diff
@@ -149,6 +149,24 @@ dotnet run
...
+```powershell
+dotnet run --devtools          # your app receives --devtools
+dotnet run -- --devtools       # identical: the SDK consumes the -- before forwarding
+```
+
+Configure the WinApp launcher itself with the `WinAppRun*` MSBuild properties. MSBuild consumes
+these, so they never reach your application:
+
+```powershell
+dotnet run -p:WinAppRunDebugOutput=true -p:WinAppRunDetach=true --devtools
```
nmetulev: This example cannot run: these properties become `--debug-output --detach`, which `RunCommand` rejects as mutually exclusive. Please remove either `WinAppRunDebugOutput` or `WinAppRunDetach` from the example.

src/winapp-CLI/WinApp.Cli/Commands/RunCommand.cs
```diff
@@ -752,6 +754,93 @@ private async Task UnregisterDevPackageAsync(string packageName, CancellationTok
...
+                if (!knownOptions.Contains(name) || !alreadyReported.Add(name))
+                {
+                    continue;
+                }
+
+                // Written through ansiConsole (gated on the logger level) to match how the rest of this
+                // command emits user-facing text, so --quiet still suppresses it.
+                var replacement = OptionToMSBuildProperty.TryGetValue(name, out var property)
+                    ? property
+                    : $"WinAppRunArgs=\"{arg}\"";
```
nmetulev: Collecting every option and alias makes legitimate application flags such as `--help`, `--configuration`, and `-p` trigger migration notices. The fallback also drops following values: for `--configuration Release` it recommends `WinAppRunArgs="--configuration"`, which then fails with a missing argument. Please limit notices to a curated set with known-valid replacements, or remove the generic fallback.

nmetulev (changes requested):

Requesting changes for three validated NuGet-path issues. The core separator change is correctly isolated: direct `winapp run`, npm-tagged calls, project mode, and `RunPackagedApp` retain their existing behavior.

src/winapp-CLI/WinApp.Cli/Commands/RunCommand.cs
azchohfi: Confirmed and fixed in `2c7e479c` — this was the worst of the three, and it shipped advice that fails.

Reproduced all of it on the built binary:

```
> winapp run . --caller nuget-package -- --help
ℹ '--help' was passed to your application, not to winapp. To configure winapp, use -p:WinAppRunArgs="--help" instead.

> winapp run . --caller nuget-package -- --configuration Release
ℹ '--configuration' was passed to your application, not to winapp. To configure winapp, use -p:WinAppRunArgs="--configuration" instead.

> winapp run . --caller nuget-package --configuration      # following that advice
❌ Required argument missing for option: '--configuration'.
```

I took the stronger of your two options and did both: `OptionToMSBuildProperty` is now the entire trigger set, and the generic fallback is gone. That fixes the two problems with one change — an option only produces a notice if there is a property that genuinely replaces it, so no suggestion can drop a value.

It's also the right rule conceptually, not just a filter: the project-mode options are *ignored in folder mode*, which is the only mode the NuGet targets use, so they never had a winapp meaning on this path. There was nothing to migrate.

After:

| forwarded | before | after |
|---|---|---|
| `--help`, `--configuration`, `-p`, `--no-build`, `--verbose` | notice | silent |
| `--detach`, `--clean`, `--executable=foo.exe` | notice | notice |

Covered by a `[DataRow]` regression test over the false-positive set, plus one asserting `--configuration Release` reaches the app with its value intact and produces no `WinAppRunArgs` suggestion.

docs/guides/dotnet.md
azchohfi: Right, and verified — fixed in `2c7e479c`.

```
> winapp run . --debug-output --detach
❌ --detach and --debug-output cannot be used together.
```

Embarrassing detail: this PR is the one that *adds* the mutually-exclusive constraint table, and the example contradicts it. Reduced to a single property:

```powershell
dotnet run -p:WinAppRunDebugOutput=true --devtools
```

I also pointed the surrounding sentence at the constraint table, so a reader combining properties is told the restrictions exist rather than discovering them at runtime.

docs/dotnet-run-support.md
azchohfi: Correct — fixed in `2c7e479c`. I verified it with a probe app that echoes its argv:

```
dotnet run --configuration Release      → APPARGS=[]
dotnet run -- --configuration Release   → APPARGS=[--configuration|Release]
```

So "optional and has no effect" was true for a token like `--devtools` and false the moment the app's flag collides with a `dotnet run` option — which is exactly when someone needs the docs to be right.

Qualified in all five places you listed (`docs/dotnet-run-support.md`, `docs/guides/dotnet.md`, `docs/usage.md`, `src/winapp-NuGet/README.md`, and the frameworks skill). Each now says the separator is optional *for arguments that do not collide with a `dotnet run` option*, names the colliding set (`--configuration`, `--framework`, `--project`, `-c`, `-f`, `-r`, ...), and shows the working form.

Nothing to fix in code here — it's the SDK's parser consuming those tokens before winapp is involved — so this stays a documentation qualification.

azchohfi:
@nmetulev — all three addressed in `2c7e479c`, each reproduced on the built binary first. Re-requested your review.

**1. Migration notices on legitimate app flags.** Took both of your options rather than either: `OptionToMSBuildProperty` is now the entire trigger set *and* the generic fallback is gone. One change fixes both halves — an option only produces a notice when a property genuinely replaces it, so no suggestion can silently drop a value.

Verified before/after:

| forwarded | before | after |
|---|---|---|
| `--help`, `--configuration`, `-p`, `--no-build`, `--verbose` | notice | silent |
| `--detach`, `--clean`, `--executable=foo.exe` | notice | notice |

Your framing was the useful part: the project-mode options are ignored in folder mode, which is the only mode the NuGet targets use, so they never had a winapp meaning on this path. Nothing to migrate. Guarded by a `[DataRow]` regression test plus one asserting `--configuration Release` reaches the app with its value intact.

**2. Impossible example.** Confirmed — `--debug-output --detach` errors as mutually exclusive, in the same PR that adds the constraint table. Reduced to one property and pointed the surrounding text at the table.

**3. `--` is not optional on collision.** Confirmed with a probe app:

```
dotnet run --configuration Release      → APPARGS=[]
dotnet run -- --configuration Release   → APPARGS=[--configuration|Release]
```

Qualified in all five places you listed, each naming the colliding option set and showing the working form. Code can't fix this one — the SDK consumes those tokens before winapp is involved — so it stays a docs qualification.

138/138 `RunCommandTests` pass; 29/29 checks green.

PR #747, opened 2026-08-13
Use the preinstalled Rust toolchain instead of an unpinned third-party action

azchohfi:
## Problem

`dtolnay/rust-toolchain@stable` was the **only** action in the repo referenced by a mutable tag instead of a commit SHA:

```yaml
- name: Setup Rust
  uses: dtolnay/rust-toolchain@stable
```

Every other action — `actions/*`, `github/codeql-action/*`, `dorny/test-reporter`, `subosito/flutter-action` — is pinned to a 40-character SHA with a version comment. So this reads as an oversight rather than a decision, and it means whoever controls that tag can run arbitrary code in our CI the next time it moves, with no review and no diff.

## Change

`windows-latest` already ships rustup with a stable toolchain, so the step can select what is on the image:

```yaml
- name: Setup Rust
  shell: pwsh
  run: |
    rustup default stable
    rustc --version
    cargo --version
```

That removes the third-party action *and* the toolchain download in one go — the second matters for the network-isolation work coming under the approved-sources requirement, since it drops a build-time fetch from `static.rust-lang.org`.

Version output is echoed so the toolchain actually used is visible in the log, which the action used to provide.

## Validation

- Workflow YAML parses (`ConvertFrom-Yaml`), jobs intact: `build`, `test-sample`, `test-samples-result`.
- No unpinned external actions remain anywhere in `.github/workflows` — verified by scanning for `uses:` refs that are not 40-char SHAs.
- The `rust-app` and `tauri-app` matrix legs exercise this; leaving CI to confirm the preinstalled toolchain resolves, since that depends on the runner image rather than anything reproducible locally.

## Context

Found while auditing for SDL requirement *"Production builds must only consume dependencies from approved sources"* (in effect 10/1/2026). This is the subset of that audit that is both a live risk today and unambiguous to fix. The remaining findings need decisions and are being tracked separately — notably `choco install innosetup` in this same workflow (Chocolatey is explicitly prohibited by that requirement), the two other third-party actions, and `networkIsolationPolicy: Permissive` in the ADO pipelines.

### DIFF
--- a/.github/workflows/test-samples.yml
+++ b/.github/workflows/test-samples.yml
@@ -154,7 +154,13 @@ jobs:
       if: >-
         steps.check.outputs.skip != 'true' &&
         contains(fromJson('["rust-app", "tauri-app"]'), matrix.sample)
-      uses: dtolnay/rust-toolchain@stable
+      shell: pwsh
+      # windows-latest already ships rustup with a stable toolchain, so this selects
+      # what is on the image rather than pulling a third-party action or fetching a toolchain.
+      run: |
+        rustup default stable
+        rustc --version
+        cargo --version
 
     # Install the Inno Setup compiler (ISCC.exe) so the sparse-app installer test
     # actually compiles installer/setup.iss instead of being skipped for a missing ISCC.


azchohfi:
Correcting the "Context" section of the description: I justified this partly as SDL approved-sources compliance. That framing was wrong and I have withdrawn it.

This repo is open source, and the GitHub Actions workflows exist so external contributors can build and pull dev builds from their PRs without Microsoft infrastructure. The approved-sources requirement explicitly anticipates that — its Microsoft-Managed OSS clause says external parties must be able to build without authenticating to an Azure Artifacts feed, and that public registries *should* be referenced in build scripts intended for external users. Production builds are the ADO pipelines in `.pipelines/`, which already source NuGet and npm exclusively from the internal feed. So nothing in this workflow was non-compliant, and the other findings I raised against these workflows (Chocolatey, `subosito/flutter-action`, `dorny/test-reporter`, crates.io, `dotnet workload`) are withdrawn.

**The change still stands on its own merits, just for a different reason.** `dtolnay/rust-toolchain@stable` was the only action in the repo on a mutable tag rather than a commit SHA. Whoever controls that tag can run arbitrary code in CI the next time it moves, with no review and no diff — that is worth closing regardless of which compliance bucket the workflow sits in, and it makes the pinning policy uniform across every workflow.

The incidental benefit of dropping a `static.rust-lang.org` fetch is real but no longer load-bearing, since these workflows are not the ones that need network isolation.

nmetulev (approved):

PR #751, opened 2026-08-14
Fix the two flaky tests taking down build-and-package

### LINKED ISSUE #729: Flaky test: RunDotnetProcessAsync_Cancellation_KillsProcessTree intermittently fails in CI
`DotNetServiceTests.RunDotnetProcessAsync_Cancellation_KillsProcessTree` (added in #686) fails intermittently in the `Build and Package` workflow on pull requests that cannot affect it.

**Source:** [src/winapp-CLI/WinApp.Cli.Tests/DotNetServiceTests.cs#L1695](https://github.com/microsoft/winappCli/blob/51666c0d/src/winapp-CLI/WinApp.Cli.Tests/DotNetServiceTests.cs#L1695)

### Observed

| PR | Scope of the change | Result |
|---|---|---|
| #726 | sample test harness only (no product code) | failed |
| #728 | two YAML config files (`dependabot.yml`, `codeql-config.yml`) | failed |
| `main` @ `51666c0d` | — | passed |

Neither PR touches anything that could influence process-tree teardown, and the same commit passed on `main`, so this looks intermittent rather than broken.

Failing runs:
- #726 - https://github.com/microsoft/winappCli/actions/runs/31525372392
- #728 - https://github.com/microsoft/winappCli/actions/runs/31525574112

### Signal

The failure takes ~15s before reporting, which suggests a timeout elapsing while waiting on child-process teardown rather than an assertion mismatch. A loaded CI runner plausibly exceeds whatever fixed wait the test uses.

Likely fix: wait deterministically on child-process exit (poll for the PIDs to disappear, or await the process handles) instead of a fixed delay/timeout.

### Possibly related

`CrashDumpServiceWorkflowTests.AnalyzeDumpAsync_RealManagedDump_RunsClrMdManagedEnumeration` (added in #644) failed in the same #728 run. Different subsystem and much older, so it may be an unrelated second flake — worth confirming before folding it in here.

azchohfi:
## Problem

The last ten `Build and Package` failures came from **two independent flaky tests**, not one. On Aug 13 the workflow failed 13 of 33 runs (~39%).

| Test | Runs | Failure |
|---|---|---|
| `DotNetServiceTests.RunDotnetProcessAsync_Cancellation_KillsProcessTree` | 6/10 | `TimeoutException: descendant PID file was not written within 00:00:15` |
| `CrashDumpServiceWorkflowTests.AnalyzeDumpAsync_RealManagedDump_RunsClrMdManagedEnumeration` | 5/10 | `StringAssert.Contains` — analysis log lacks `CLR Version` |

### 1. Tree-kill — #732 fixed a real bug, just not this one

#732 and #737 fixed a **sharing-violation `IOException`** in `WaitForPidFileAsync`, and that signature is gone. The current failure is a different mode. It already appeared on Aug 12 at 17:39, *before* #737 merged, and 4 of the 6 failures are on commits that **contain** the fix — including one on `main`. Issue #729 is still open.

Root cause, reproduced directly: `Start-Process` failure is a **non-terminating** error, so the script ran on to `Set-Content` with `$p` still null and published a **0-byte file**:

```
File created: YES; length=0 bytes; content=<>; int.TryParse=False
```

`int.TryParse` silently skipped it and the reader polled the full 15s. The message "was not written" was actively misleading — the file *was* written, just empty. The child's stderr was redirected and discarded, so the real error never reached CI. For scale, that script publishes in **373 ms** on an unloaded machine, against a 15 s budget.

### 2. Managed dump — never fixed at all

`TestProcessDump.TryCreateManagedDump` dumped a PowerShell child after a blind `Thread.Sleep(1500)`. Under load the child was still in **native** startup — the CI dumps show only `hostfxr`/`hostpolicy` loading coreclr — so ClrMD found no runtime. No commit had ever touched this.

## Fixes

**Tree-kill**
- `$ErrorActionPreference = 'Stop'` plus an explicit null guard, so a failed child start aborts *before* `Set-Content` rather than publishing an empty file
- `WaitForPidFileAsync` now watches the launcher task and fails fast with the root's exit code and stderr instead of waiting out the clock
- the timeout message distinguishes never-created, unreadable, and created-but-unparsable

**Managed dump**
- the child allocates, forces a collection, and only then signals readiness, so the dump is taken with the runtime up **and** its heap walkable

## On the second fix, specifically

Signalling readiness alone was **not** enough, and this is worth naming because the obvious fix is wrong. A token emitted from the first line of managed code proves the CLR started but not that ClrMD can walk it — that version made the dump test fail or go inconclusive on a machine where the old sleep passed 6/6. Warming the heap before signalling is what actually fixes it:

| Version | Idle | Under 100% CPU load |
|---|---|---|
| Original (`Thread.Sleep(1500)`) | 6/6 pass | 6/6 pass |
| Token only | **3/6 pass** | — |
| Token after heap warm-up | 8/8 pass | 6/6 pass |

## Verification

- Both regression guards were checked against the **original** code, not just the new code. `TreeKillScript_WhenTheChildCannotStart_PublishesNoPidFileAtAll` fails with the original script (`Assert.IsFalse failed ... File.Exists(pidFile)`) and passes with the hardened one.
- The script builder is shared between the tree-kill test and its regression test, so the script proven not to publish an empty file is the one actually run — the two cannot drift.
- Full suite: **4,572 passed / 4,581**. The 4 failures are `winapp init` E2E tests, confirmed pre-existing by re-running them with these changes stashed — they fail on `Failed to get version for Microsoft.WindowsAppSDK`, the documented corp-machine nuget.org restriction.

Local load testing could not reproduce CI's 4-core I/O-bound conditions, so the load column above is weaker evidence than the idle column. The structural argument stands on its own: the old code bounded CLR startup with a fixed 1.5 s guess, and CI captured dumps proving that guess was exceeded.

Closes #729

### DIFF
--- a/src/winapp-CLI/WinApp.Cli.Tests/DotNetServiceTests.cs
+++ b/src/winapp-CLI/WinApp.Cli.Tests/DotNetServiceTests.cs
@@ -1704,13 +1704,7 @@ public class DotNetServiceTests : BaseCommandTests
         // with -PassThru, writes ping's PID to a temp file, then blocks on Wait-Process. The test polls
         // that file — no WMI or Win32 process enumeration required.
         var pidFile = Path.Join(Path.GetTempPath(), $"winapp_treekill_{Guid.NewGuid():N}.pid");
-        // Escape single quotes for the single-quoted PowerShell string literal so a temp path
-        // containing an apostrophe (e.g. a profile like O'Brien) still produces a valid script.
-        var pidFileLiteral = pidFile.Replace("'", "''");
-        var script =
-            "$p = Start-Process -FilePath ping.exe -ArgumentList '-n','60','127.0.0.1' -PassThru -WindowStyle Hidden; " +
-            $"Set-Content -LiteralPath '{pidFileLiteral}' -Value $p.Id; " +
-            "Wait-Process -Id $p.Id";
+        var script = BuildTreeKillScript(pidFile, "ping.exe");
 
         var startInfo = new ProcessStartInfo
         {
@@ -1744,8 +1738,11 @@ public class DotNetServiceTests : BaseCommandTests
             rootPid = rootPidValue;
             Assert.IsFalse(Process.GetProcessById(rootPidValue).HasExited, "The root process should be running before cancellation.");
 
-            // Capture the ping descendant via the PID file the root script writes.
-            var pingPidValue = await WaitForPidFileAsync(pidFile, TimeSpan.FromSeconds(15), TestContext.CancellationToken);
+            // Capture the ping descendant via the PID file the root script writes. The run task is passed
+            // in so a root that dies before publishing fails here with its own stderr rather than after a
+            // silent wait. The budget is generous because it costs nothing on success (a healthy spawn
+            // publishes in well under a second) and only bounds a genuinely wedged agent.
+            var pingPidValue = await WaitForPidFileAsync(pidFile, TimeSpan.FromSeconds(30), TestContext.CancellationToken, runTask);
             pingPid = pingPidValue;
             Assert.IsFalse(Process.GetProcessById(pingPidValue).HasExited, "The ping descendant should be running before cancellation.");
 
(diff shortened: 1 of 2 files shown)


zateutsch (approved):
PR review skill was all clean, looks good to me.

PR #756, opened 2026-08-14
Update winapp new to warn when scaffolding is taking longer than expected

### LINKED ISSUE #752: [Feature]: Detect when WinUI templates need to resolve WASDK and display loading text
The WinUI templates dynamically resolve their WASDK version, and may need to download latest if there has been an update since last run. This can cause the new command to run for much longer than expected on first run.

We don't have a way around this, but we should inform the user of what is happening so there's no confusion.

zateutsch:
## Summary
 
 Updates `winapp new` to provide clearer feedback when project scaffolding takes longer than expected.
 
 - Keeps the existing scaffold spinner for the first 10 seconds.
 - If scaffolding is still running, changes the status to:
   > Still setting up the project; missing NuGet packages may be downloading or restoring…
 - Uses elapsed time rather than parsing `dotnet new` output, preserving localized verbose and error messages.
 - Leaves `--json`, `--quiet`, redirected output, CI, and other non-interactive behavior unchanged.
 
 Fixes #752.
 
 ## Testing
 
 - Built `src\winapp-CLI\winapp.sln` in Release with 0 warnings and 0 errors.
 - Ran 139 focused `NewCommand` tests successfully.
 - Published the ARM64 CLI and used the published binary to create a real WinUI project with the installed template pack.
 - Ran the PR review skill; no critical or high findings remained after validation and gut check.


src/winapp-CLI/WinApp.Cli/Commands/NewCommand.cs
```diff
@@ -681,6 +681,8 @@ private async Task<int> InvokeCoreAsync(ParseResult parseResult, InvocationTelem
             var verb = entry.IsItem ? "Adding" : "Creating";
             var (exitCode, stdout, stderr) = await WithSpinnerAsync(
                 $"{verb} {name} from {entry.ShortName}...",
+                "Still setting up the project; missing NuGet packages may be downloading or restoring…",
```
nmetulev:
```suggestion
                "Setting up the project; missing NuGet packages are restoring…",
```

nmetulev (approved):

PR #766, opened 2026-08-18
Publish release symbols to symweb and MSDL

azchohfi:
Publishes the release PDBs so a crash dump or hang from a shipped build can actually be resolved to a function name - to the internal symbol server (symweb) for full private symbols, and to the public server (MSDL) as stripped public symbols. Today the only PDBs anywhere are the ones inside the public GitHub release zips, which is not somewhere a debugger will look.

### What this adds

A `Publish_Symbols` stage to the release pipeline, gated on a new `PublishSymbols` parameter (default `true`, so it can be turned off for a re-run without editing the pipeline). It runs off the existing `cli-binaries` artifact and does three things:

1. Acquires a token for the symbol request API via the `WinAppCLISymbolPublishing` service connection.
2. Uploads the PDBs with `PublishSymbols@2`.
3. Registers a symbol request and promotes it to both servers.

### On the public server

`publishToPublicServer` is `true`. Worth being precise about what does and does not go public, because it is easy to assume this leaks private symbols:

The symbol project has no allowed-private-symbols policy, and that policy is *not* a gate on public publishing - it only decides whether private symbols are published to MSDL **unstripped**. With no policy, `OrchestratorUtils.CreateSymbolFileStrippingInformation` sets `stripSymbolFile = true` for any non-public PDB bound for the external repo, so the service strips the private data before it goes out. What lands on MSDL is public symbols: enough to resolve a stack to function names, without source, lines, locals, or types.

So this is a strict improvement in debuggability with no new disclosure. (Full private PDBs are already inside the public release zips today, so even the stripped-vs-full distinction is not the exposure boundary it would normally be.)

### Notes for review

**Expiration is ~100 years.** NativeAOT PDBs are large, but they are the only way to debug a shipped build, and a request that quietly expires is worse than one that never existed. Note the widely-copied `Create-SymbolPublishingRequest.ps1` sample defaults to *6 days*.

**`expirationTime`, not `expirationDate`.** That same sample sends `expirationDate`. The service contract (`RegisterSymbolRequest`) declares only `RequestName` and `ExpirationTime`, and Newtonsoft drops unknown members - so the sample's expiration parameter is silently a no-op and the request falls back to the 30-year default. Please do not "fix" this field to match the sample.

**The date is round-trip UTC** (`ToString('o')`) rather than the bare `.ToString()` some existing pipelines use, since the service binds it to a `DateTime` and a culture-formatted string would be ambiguous on a non-en-US agent.

**`SymbolPublishingProject` is case sensitive** and must match the provisioned project exactly or the request 404s.

**`SymbolsArtifactName` is used twice** - as the `PublishSymbols@2` artifact name and as the request name being promoted. They must stay identical or the promote step finds nothing to publish.

`IndexSources` is off. Source indexing does not map cleanly onto a GitHub-sourced NativeAOT build, and it is not required here.

### Testing

A pipeline stage cannot be fully exercised until it runs on a real release. The YAML parses and the stage nests as a sibling of the existing conditional release stages. The request field name, the stripping behaviour on the public path, and the absence of a retention cap were each confirmed against the symbol service's own source rather than against sample scripts or docs.

### DIFF
--- a/.pipelines/release.yml
+++ b/.pipelines/release.yml
@@ -311,7 +311,11 @@ extends:
     - ${{ if eq(parameters.PublishSymbols, 'true') }}:
       - stage: Publish_Symbols
         displayName: Publish Symbols
-        dependsOn: Build
+        # Build supplies the version output; Release_GitHub gates on the release actually shipping, so
+        # a failed release cannot leave century-retention symbols published for a build nobody has.
+        dependsOn:
+          - Build
+          - Release_GitHub
         variables:
           version: $[ stageDependencies.Build.Build.outputs['SetMeta.version'] ]
           # Used twice: as the PublishSymbols@2 artifact name and as the request name the symbol API
@@ -349,6 +353,9 @@ extends:
             # why the symbol project's source organization must be `microsoft` rather than the default.
             - task: PublishSymbols@2
               displayName: Upload PDBs to the Azure DevOps symbol service
+              # The artifact name is fixed per build, so a re-run after a later failure would fail here
+              # on the existing request and never reach the REST calls. Surfaces as a warning instead.
+              continueOnError: true
               inputs:
                 SymbolsFolder: $(Pipeline.Workspace)/symbolStaging
                 SearchPattern: '**/*.pdb'
@@ -385,14 +392,14 @@ extends:
                 Invoke-RestMethod -Uri $baseUri -Body ($createBody | ConvertTo-Json -Compress)
 
                 # Public symbols land on MSDL stripped: with no allowed-private-symbols policy on the
-                # project, the service runs the private data out of the PDB before it goes external.
+                # project, the service strips the private data out of the PDB before it goes external.
                 $publishBody = @{
                     publishToInternalServer = $true
                     publishToPublicServer   = $true
                 }
                 Write-Host "Promoting $($createBody.requestName) to symweb and MSDL"
                 Invoke-RestMethod -Uri "$baseUri/$($createBody.requestName)" -Body ($publishBody | ConvertTo-Json -Compress)
-              displayName: Request publication to symweb
+              displayName: Request publication to symweb and MSDL
 
     - ${{ if eq(parameters.DoEsrp, 'true') }}:
       - stage: Release_Npm
--- a/.pipelines/release.yml
+++ b/.pipelines/release.yml
@@ -37,6 +37,11 @@ parameters:
     displayName: Publish symbols to symweb and MSDL (PublishSymbols)
     type: boolean
     default: true
+  - name: symbolsIdentity
+    type: object
+    default:
+      serviceConnection: $(SymbolPublishingServiceConnection)
+      project: $(SymbolPublishingProject)
 
 variables:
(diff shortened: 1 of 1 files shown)


azchohfi:
Updated: the symbol service connection and project name are no longer literals in the YAML.

This repo is public, so `release.yml` was publishing internal ADO infrastructure names. They now come from pipeline variables through a `symbolsIdentity` parameter object, which is the same indirection the existing ESRP `signingIdentity` parameter already uses:

```yaml
- name: symbolsIdentity
  type: object
  default:
    serviceConnection: $(SymbolPublishingServiceConnection)
    project: $(SymbolPublishingProject)
```

**Two variables need to exist on the pipeline (or in a variable group) before this stage can run:** `SymbolPublishingServiceConnection` and `SymbolPublishingProject`. The stage fails at the token step if they are missing.

### On marking them secret

Worth being deliberate here rather than defaulting to "secret is safer". The existing `signingIdentity` variable group splits exactly along this line:

| Variable | isSecret |
|---|---|
| `SigningAppId`, `SigningTenantId`, `SigningAKVName`, cert names | `true` |
| `SigningServiceName` | `false` |

The service connection *name* is deliberately not secret there, while everything genuinely sensitive is. That split looks intentional - a service connection name is a reference, not a credential; the actual secret material lives in the connection and is protected by pipeline authorization, and ADO resolves connection authorization from that input before the job runs.

So the recommendation is to add both as **non-secret** variables in a variable group. The disclosure concern is that these names sat in a public repo, and moving them into ADO already solves that. I have not verified that a secret-valued `azureSubscription` input resolves correctly, and the release pipeline is not somewhere I would want to discover that it does not - if we do want them secret, that is worth confirming on a throwaway run first.

azchohfi:
### Correction: the service connection name I gave earlier was wrong

My earlier comment said to create `SymbolPublishingServiceConnection` with the value `WinAppCLISymbolPublishing`. **That name does not exist.** The real connection in `microsoft/pde-oss` is:

```
WinAppCLI - Symbol Publishing
```

(spaces and a hyphen; endpoint id `88f9f214-e52a-44d5-a77c-5394c395847c`, azurerm / WorkloadIdentityFederation). Confirmed against `_apis/serviceendpoint/endpoints`. The old value would have failed the token step on the first real release.

This is a good argument for the indirection itself - the wrong name was previously baked into the YAML, where it would only have been discovered during a release.

### The connection is not authorized for the release pipeline

`pipelinePermissions/endpoint/88f9f214-...` currently returns:

```json
{ "resource": { "type": "endpoint", "id": "88f9f214-..." }, "pipelines": [] }
```

Authorized for nothing. For comparison, the ESRP connection is explicitly authorized for `182109` (and six other definitions), which is why the signing stage works.

This matters more than usual here: because the connection name arrives through a variable, ADO cannot resolve it statically, so it will not surface the usual "grant permission" prompt at queue time. It has to be granted up front or the job fails.

**Needed:** authorize `WinAppCLI - Symbol Publishing` for pipeline `182109` (WinDevCLI - Release), or set "Grant access permission to all pipelines" on the connection.

nmetulev (approved):

PR #789, opened 2026-08-26
Stream project restore progress during winapp run

nmetulev:
## Summary

- stream pre-build `dotnet restore` output, using dotnet's native terminal logger when interactive and raw line streaming when redirected
- keep `--quiet` restores quiet, preserve JSON stdout, redact displayed commands, and avoid duplicate verbose command output
- omit resolved `Platform` from solution-scoped restores so configuration-free `.slnx` files do not emit `MSB4126`
- explain when best-effort restore failures fall back or continue, and preserve subprocess line formatting including blank lines
- document the output behavior and add real `.slnx` sample coverage

## Validation

- 199 `ProjectRunServiceTests` passed in Release
- real configuration-free `WinUISolution.slnx` run restored, built, and launched without `MSB4126`; the solution restore omitted `-p:Platform` while the project build retained it
- plugin package validation and `git diff --check` passed

## Environment note

The complete `scripts\build-cli.ps1` run reached NativeAOT publish but could not finish on this device because the Visual Studio Desktop Development for C++ linker workload is not installed.


src/winapp-CLI/WinApp.Cli/Services/ProjectRunService.cs
```diff
@@ -602,14 +584,64 @@ private async Task RestoreSiblingsIndividuallyAsync(
...
+            {
+                Console.Error.WriteLine($"dotnet {RedactSecretsForDisplay(arguments)}");
+            }
+
+            return await dotNetService.RunDotnetStreamingAsync(
+                workingDir, arguments,
+                onOutputLine: static line => Console.Error.WriteLine(line),
+                onErrorLine: static line => Console.Error.WriteLine(line),
+                cancellationToken);
+        }
```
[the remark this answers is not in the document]
nmetulev: Fixed in the stacked follow-up PR #791. Quiet-mode restores now receive `-v quiet`, with direct argument-builder, routing, and full pipeline regression coverage. The Release `ProjectRunServiceTests` pass 195/195, and a real throwaway-solution run confirmed quiet mode emits neither the restore invocation nor normal restore progress.

docs/usage.md
```diff
@@ -754,15 +754,15 @@ Project mode requires the **.NET SDK 8.0.100 or newer** (for MSBuild `--getPrope
...
+**Build output & verbosity:** dependency restore and `dotnet build` output **stream live** to your console, followed by a fast property-evaluation pass. In an interactive terminal, dotnet's terminal logger shows in-place progress and elapsed time. For redirected output and CI, winapp streams plain lines instead. It prints each exact `dotnet restore …` or `dotnet build …` invocation first, so package downloads, feed retries, errors, and build warnings remain visible. Verbosity:
 
 | Flag | dotnet verbosity | Adds |
 |------|------------------|------|
 | *(default)* | `minimal` | — |
 | `--verbose` | `minimal` | winapp's build decision traces |
 | `--quiet` | `quiet` | — |
 
-Under `--json` or `--quiet` the invocation and build output go to stderr so stdout stays pure JSON / clean.
+Under `--json` or `--quiet` the invocation, restore output, and build output go to stderr so stdout stays pure JSON / clean.
```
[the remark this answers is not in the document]
nmetulev: Fixed in the stacked follow-up PR #791. The usage reference now distinguishes the modes accurately: `--json` sends the invocation and child output to stderr, while `--quiet` suppresses the invocation and sends quiet child output to stderr.

src/winapp-CLI/WinApp.Cli/Services/ProjectRunService.cs
```diff
@@ -569,7 +550,8 @@ private async Task<bool> RestoreSolutionSiblingsAsync(FileInfo target, ProjectRu
             // Closest to VS: one restore over the whole solution pulls the target and every sibling.
             var args = BuildRestorePassArguments(options.Solution, options);
             logger.LogDebug("{UISymbol} Restoring solution before build (build-dependency parity): dotnet {Arguments}", UiSymbols.Note, RedactSecretsForDisplay(args));
-            var (exitCode, _, _) = await dotNetService.RunDotnetCommandAsync(workingDir, args, cancellationToken);
+            var exitCode = await RunRestoreCommandAsync(
```
azchohfi, line 560: This turns a deliberately best-effort restore into visible error output, and on `.slnx` solutions that means a red MSBuild error on runs that actually succeed.

Repro — `dotnet new reactorapp` + `dotnet new classlib` + `dotnet new sln` + `dotnet sln add`, then `winapp run . --arch x64`:

```
🔄 Restoring ReactorSln.slnx dependencies...
ReactorSln.slnx.metaproj : error MSB4126: The specified solution configuration "Debug|x64" is invalid.
...
Build succeeded.
✅ Built ReactorRunTest in 4.0s
```

`main` on the same input prints nothing extra. The old `var (exitCode, _, _) = ...` discarded the output, so this restore has always been failing here — it was just invisible. We recover from it (fall through to per-sibling restore), so the user now sees a hard-looking error for something we intentionally handle.

Root cause is `BuildRestorePassArguments` injecting `-p:Platform=x64` on a *solution* target. MSBuild treats Platform as the solution configuration there, and `.slnx` declares none:

```
dotnet restore App.slnx -r win-x64 -p:Configuration=Debug -p:Platform=x64   → error MSB4126
dotnet restore App.slnx -r win-x64 -p:Configuration=Debug                   → Restore succeeded
```

Not a corner case: `dotnet new sln` emits `.slnx` by default on SDK 10.0.400, and every WinUI/Reactor template declares `<Platforms>`, which is exactly what triggers the injection.

Smallest fix: don't inject `-p:Platform` for the solution-scoped restore. The per-project sibling restores below can keep it.

src/winapp-CLI/WinApp.Cli/Services/ProjectRunService.cs
```diff
@@ -602,14 +584,64 @@ private async Task RestoreSiblingsIndividuallyAsync(
...
+    /// render in-place progress; redirected, JSON, and quiet modes use line streaming instead.
+    /// </summary>
+    private async Task<int> RunRestoreCommandAsync(
+        string arguments,
+        string banner,
+        ProjectRunOptions options,
+        DirectoryInfo workingDir,
+        CancellationToken cancellationToken)
+    {
+        if (options.Json || !logger.IsEnabled(LogLevel.Information))
```
azchohfi, line 635: `--quiet` gets louder with this change. The build pass runs `ResolveBuildVerbosity` and hands dotnet `-v quiet`, but `BuildRestorePassArguments` never passes `-v` at all, so the restore runs at dotnet's default `minimal` and streams straight to stderr.

`winapp run <sln> --quiet`, before vs after:

```
# main
Build succeeded.

# this branch
  Determining projects to restore...
  Restored Lib.csproj (in 194 ms).
  Restored ReactorRunTest.csproj (in 314 ms).
Build succeeded.
```

Fix: thread the same resolved verbosity into the restore args so a `--quiet` run stays quiet.

src/winapp-CLI/WinApp.Cli.Tests/FakeDotNetService.cs
```diff
@@ -133,6 +133,19 @@ public Task<int> RunDotnetStreamingAsync(DirectoryInfo workingDirectory, string
         {
             return Task.FromResult(RunDotnetStreamingHandler(arguments, onOutputLine, onErrorLine));
         }
+        if (RunDotnetCommandHandler is not null)
+        {
+            var (exitCode, output, error) = RunDotnetCommandHandler(arguments);
+            foreach (var line in output.Split(['\r', '\n'], StringSplitOptions.RemoveEmptyEntries))
```
azchohfi: Non-blocking, but worth knowing this doesn't match the real service: `RunDotnetCoreAsync` forwards `e.Data` whenever it's non-null, so genuine blank lines reach the callbacks as `""`. Here they get dropped.

Harmless today — both callbacks only write lines, nothing parses or counts them — and `RemoveEmptyEntries` is effectively required given the split is on the individual `'\r'`/`'\n'` chars (without it every CRLF would emit a bogus empty entry, which would be less faithful, not more). Just means a future blank-line regression wouldn't be caught here.

src/winapp-CLI/WinApp.Cli.Tests/ProjectRunServiceTests.cs
```diff
@@ -2871,20 +2871,57 @@ public async Task BuildAndResolveAsync_SolutionAllManaged_RestoresWholeSolutionT
...
+        StringAssert.Contains(console.Output, "Restoring App.slnx dependencies",
+            "the restore phase should be announced before dotnet starts");
+        StringAssert.Contains(console.Output, "RESTORE-PROGRESS",
+            "restore output should stream live instead of being buffered until dotnet exits");
+        StringAssert.Contains(dotnet.StreamingCalls.Single(a => a.StartsWith("build ", StringComparison.Ordinal)), "--no-restore",
             "the build pass should skip its own restore since the solution restore already covered the target");
     }
 
+    [TestMethod]
+    public async Task BuildAndResolveAsync_SolutionRestoreInRealTerminal_UsesInheritedLauncher()
```
azchohfi: These new tests use `.slnx` fixtures, but `FakeDotNetService` can't reproduce MSBuild's solution-configuration validation, so the `-p:Platform` + `.slnx` failure I hit in real usage still passes here. There's also no `.slnx` anywhere in `samples/` — only `WinUISolution.sln` — which is why all 30 checks are green despite the MSB4126 error above.

Worth adding a `.slnx` case to the sample tests, since that's the format `dotnet new sln` produces by default now.

azchohfi (dismissed):
The direction here is right, and on a classic `.sln` this works exactly as intended — restore progress streams, `--no-restore` is set correctly on the build, and `--json` stdout stays pure.

One blocker: on `.slnx` solutions this now surfaces an `error MSB4126` on runs that succeed. It's a pre-existing latent failure that the old buffered call was hiding, so the streaming change is doing its job — but the failure itself needs fixing rather than displaying, since `.slnx` is what `dotnet new sln` produces by default on .NET 10 and every WinUI/Reactor template trips the condition. Found it running a Microsoft.UI.Reactor app in a two-project solution.

Also a smaller one: `--quiet` isn't quiet anymore for the restore pass.

zateutsch:

# PR Review — nmetulev-improve-run-progress vs origin/main

## Decision
**Changes required** — the restore-progress change creates two user-visible regressions: `--quiet` now emits informational restore chatter, and best-effort restore failures print alarming NuGet errors even when the app subsequently builds and launches successfully.

## Must fix
### Quiet mode emits normal restore progress
- **What is wrong:** Restore arguments do not apply quiet verbosity, so the new streaming path forwards informational `dotnet restore` output under `--quiet`.
- **Show me:** `winapp run Review.slnx --project App --quiet -p WindowsPackageType=None` → stderr contains `Determining projects to restore...` and `Restored ...`; expected: warnings and errors only.
- **Why it matters:** Quiet automation becomes noisy and violates the documented Warning-level behavior.
- **Smallest fix:** Add `-v quiet` to restore invocations in quiet mode while retaining stderr routing for warnings and errors; add a focused argument/output test.
- **Location:** `src/winapp-CLI/WinApp.Cli/Services/ProjectRunService.Arguments.cs:23-58`, `src/winapp-CLI/WinApp.Cli/Services/ProjectRunService.cs:608-619`

### Ignored restore failures look fatal
- **What is wrong:** Solution and sibling restores are explicitly best-effort, but their raw NuGet errors now stream without a visible message that winapp is continuing. A failed solution restore can also be followed by a failed sibling restore, printing the errors twice before a successful build.
- **Show me:** Run a solution whose non-target sibling references an unavailable package → winapp prints `error NU1101`, retries the sibling and prints another `error NU1101`, then reports `Build succeeded`, launches the app, and exits 0.
- **Why it matters:** Users see apparent fatal errors and may stop or troubleshoot a feed problem that did not block their requested app.
- **Smallest fix:** After each nonzero best-effort restore, print a clear visible message that winapp is continuing and that the target build will report any blocking error; keep the existing fallback behavior.
- **Location:** `src/winapp-CLI/WinApp.Cli/Services/ProjectRunService.cs:542-568`, `src/winapp-CLI/WinApp.Cli/Services/ProjectRunService.cs:577-594`

## Non-blocking
### Quiet-mode documentation promises suppressed command lines
- **What is wrong:** The usage page says quiet mode prints each restore/build invocation to stderr, but the implementation suppresses those invocations and only streams command output.
- **Show me:** `winapp run App.slnx --quiet` → no `dotnet restore ...` or `dotnet build ...` line; expected from the documentation: each invocation appears first.
- **Why it matters:** Users cannot rely on the documented quiet output when reproducing a failure.
- **Smallest fix:** Clarify that quiet mode suppresses invocations while routing restore/build output to stderr; only JSON mode prints the invocation there.
- **Location:** `docs/usage.md:757-765`

### Redirected restore output is wrapped instead of streamed as plain lines
- **What is wrong:** `ansiConsole.WriteLine` wraps redirected restore output to Spectre.Console’s profile width.
- **Show me:** Redirect a restore containing a long path or feed URL → one dotnet diagnostic is split across multiple roughly 80-column lines; expected: the original append-only dotnet line.
- **Why it matters:** CI parsers can miss diagnostics, and paths and URLs become difficult to copy.
- **Smallest fix:** Under the existing write lock, write through `ansiConsole.Profile.Out.Writer.WriteLine(line)` as the build path already does.
- **Location:** `src/winapp-CLI/WinApp.Cli/Services/ProjectRunService.cs:632-642`

## What was exercised
- `dotnet run --project src\winapp-CLI\WinApp.Cli.Tests\WinApp.Cli.Tests.csproj -c Debug -- --filter "FullyQualifiedName~ProjectRunServiceTests"` — 191/191 passed.
- NativeAOT publish for `win-arm64` and direct invocation of the published `winapp.exe` — succeeded.
- Quiet-mode run against a cold-cache two-project solution — exited 0 and reproduced informational restore output on stderr.
- Best-effort failure run with an unavailable sibling package and isolated empty feed — printed repeated NU1101 errors, then built, launched, and exited 0.
- Property command-injection attempt containing `& cmd /c` — no marker file was created.
- `.\scripts\build-cli.ps1 -SkipTests` — native x64/arm64 builds and npm/NuGet packaging completed; generated files were left unchanged.
- Not exercised: MSIX packaging — the build-tools download failed with the known corporate-network SSL restriction, which does not affect these project-run findings.

zateutsch (changes requested):
Review attached above.

src/winapp-CLI/WinApp.Cli.Tests/FakeDotNetService.cs
nmetulev: Fixed in e3571e16. The fake now forwards lines through `StringReader.ReadLine()`, preserving genuine blank lines without inventing an empty line for each CRLF separator. The restore streaming test covers consecutive blank output lines.

src/winapp-CLI/WinApp.Cli.Tests/ProjectRunServiceTests.cs
nmetulev: Fixed in e3571e16. `samples/winui-solution` now includes a configuration-free `.slnx`; Phase 1 runs solution mode against it and explicitly selects the platform-aware test project to assert real MSBuild output contains no MSB4126. The argument test separately proves solution restore omits `-p:Platform`.

src/winapp-CLI/WinApp.Cli/Services/ProjectRunService.cs
nmetulev: Fixed in e3571e16. Solution-scoped restores now omit the resolved `-p:Platform`, while project and sibling restores retain it. I added focused argument coverage plus a configuration-free `WinUISolution.slnx` sample case; a real run restored the solution without MSB4126 and built `App.Tests` with `-p:Platform=ARM64`.

src/winapp-CLI/WinApp.Cli/Services/ProjectRunService.cs
nmetulev: Fixed in e3571e16. The lower PR now includes the quiet restore routing from #791: non-JSON information-disabled restores receive `-v quiet`, suppress the invocation, and keep warnings/errors on stderr. Direct and full-pipeline regression tests cover it.

nmetulev:
@zateutsch Addressed the review findings in the latest commits: quiet restores now use `-v quiet`; every nonzero best-effort restore now prints an explicit fallback/continuation warning; redirected restore and build lines bypass Spectre wrapping; and the usage/setup docs reflect the actual JSON and quiet contracts. The branch also fixes the `.slnx` Platform failure and adds real configuration-free `.slnx` coverage. Release `ProjectRunServiceTests` pass 199/199.

src/winapp-CLI/WinApp.Cli.Tests/ProjectRunServiceTests.cs
```diff
@@ -2945,6 +3187,40 @@ public async Task BuildAndResolveAsync_WholeSolutionRestoreFails_FallsBackToPerS
...
+            "App.slnx",
+            SlnxListing("App.csproj", "Managed/Managed.csproj", "Native/Native.vcxproj"));
+        var dotnet = new FakeDotNetService
+        {
+            RunDotnetCommandHandler = args =>
+                args.StartsWith("restore ", StringComparison.Ordinal)
+                    ? (1, string.Empty, "simulated sibling restore failure")
+                    : (0, PackagedPropertiesJson(), string.Empty),
+        };
+        var console = new TestConsole();
```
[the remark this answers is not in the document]
nmetulev: Fixed in 5877debb. The sibling restore continuation test now disposes its local `TestConsole` with `using var`.

src/winapp-CLI/WinApp.Cli.Tests/ProjectRunServiceTests.cs
```diff
@@ -2936,7 +3175,10 @@ public async Task BuildAndResolveAsync_WholeSolutionRestoreFails_FallsBackToPerS
                 return (0, PackagedPropertiesJson(), string.Empty);
             },
         };
-        var service = NewServiceWith(dotnet, out _);
+        var console = new TestConsole();
```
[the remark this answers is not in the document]
nmetulev: Fixed in 5877debb. The whole-solution restore fallback test now disposes its local `TestConsole` with `using var`.

src/winapp-CLI/WinApp.Cli.Tests/ProjectRunServiceTests.cs
```diff
@@ -1942,13 +1942,40 @@ public async Task BuildAndResolveAsync_SdkAbsentShimInitiallyNull_RestoresThenRe
...
+    {
+        var csproj = WriteFile("App.csproj", ExecutableCsproj);
+        var dotnet = new FakeDotNetService
+        {
+            RunDotnetCommandHandler = args => args.StartsWith("restore ", StringComparison.Ordinal)
+                ? (1, string.Empty, "simulated restore failure")
+                : (0, PackagedPropertiesJson(), string.Empty),
+        };
+        var shim = new FakeCsWinRTMetadataShimService { WindowsSdkAbsent = true, FolderToReturn = null };
+        var console = new TestConsole();
```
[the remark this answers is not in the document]
nmetulev: Fixed in 5877debb. The test now declares the local `TestConsole` with `using var`, preserving behavior while disposing it on every exit path.

zateutsch:

# PR Review — nmetulev-improve-run-progress vs origin/main

## Decision
**Changes required** — the new whole-solution restore can produce assets for a different MSBuild platform than the subsequent `--no-restore` project build. The branch otherwise has appropriate scope, documentation, tests, and security handling.

## Must fix
### Platform-less solution restore can poison the no-restore build
- **What is wrong:** After resolving `Platform=x64` or `ARM64`, winapp restores the solution without that platform, then builds the selected project with the platform and `--no-restore`. Platform-conditioned packages or runtime identifiers are therefore absent from `project.assets.json`.
- **Show me:** A solution containing a project with an x64-only `PackageReference`: `winapp run App.slnx --arch x64 --detach` → solution restore omits `-p:Platform=x64`, then project build adds it with `--no-restore` and fails with `CS0103` or `NETSDK1047`; expected: the selected project restores under the same platform it builds with.
- **Why it matters:** Valid WinUI solutions that worked before this branch can fail with misleading missing-package or missing-assets errors.
- **Smallest fix:** Keep omitting `Platform` from solution restores to avoid `MSB4126`, but only set the selected build to `NoRestore` when no platform was resolved. When a platform exists, let the project build perform its normal project-scoped restore.
- **Location:** `src/winapp-CLI/WinApp.Cli/Services/ProjectRunService.cs:181`

## Non-blocking
### Quiet fallback warnings contradict the clean-stdout contract
- **What is wrong:** The three new restore fallback messages use `LogWarning`; this logger routes warnings to stdout even under `--quiet`, while the updated documentation promises clean stdout.
- **Show me:** `winapp run App.slnx --quiet --detach -p Platform=x64` with a failing solution restore → stdout contains the fallback warning before the detached PID; expected: restore diagnostics remain on stderr.
- **Why it matters:** Scripts capturing quiet stdout can receive an unexpected human-readable line.
- **Smallest fix:** Write these restore-failure explanations to stderr in quiet mode while retaining normal warning behavior otherwise.
- **Location:** `src/winapp-CLI/WinApp.Cli/Services/ProjectRunService.cs:518,566,597`

## What was exercised
- `scripts\build-cli.ps1` — native and npm builds succeeded; the full run reached 4,635 tests, with 70 unrelated failures from missing internal NuGet-feed configuration and one ARM64 crash-dump fixture mismatch.
- `dotnet run --project src\winapp-CLI\WinApp.Cli.Tests\WinApp.Cli.Tests.csproj -c Debug --no-build -- --filter "FullyQualifiedName~ProjectRunServiceTests"` — 199 passed, including JSON secret-redaction and stdout/stderr routing coverage.
- NativeAOT `dotnet publish` plus direct `winapp.exe --version` — publish succeeded and the binary ran.
- Synthetic `.slnx` and classic `.sln` projects — reproduced the platform-conditioned package/RID failure; allowing the selected project to restore fixed it, while forwarding `Platform` to either solution format reproduced `MSB4126`.
- Security red-team path: JSON restore with `NuGetApiKey=top-secret` — stderr contained `NuGetApiKey=***`, stdout stayed clean, and the secret was absent.
- Not exercised: the full WinUI sample Pester workflow — the focused runtime reproductions covered the changed restore behavior, but not the sample’s complete packaging flow.

zateutsch:
## PR Review — multi-dimensional

**Decision: changes required.** The core change — streaming restore output live with feed-credential redaction, plus the MSB4126 fix for configuration-free `.slnx` — is well-built, well-tested (203 ProjectRunService tests pass locally), and appropriately scoped. Two real gaps in the same feature it touches should be fixed first.

### Must fix

**1. Build-triggered restore can still print feed credentials the PR set out to redact**
- **What is wrong:** The PR routes every *restore* line through `NugetErrorMessage.Redact`, but `dotnet build` performs an implicit restore whenever `--no-restore` wasn't added, and the build pass doesn't redact. Under `--json`/`--quiet`, build child output goes to `Console.Error.WriteLine(line)` unredacted; on a native interactive terminal the build runs with inherited stdio, so winapp never sees the lines. The PR *newly* creates a path that makes the build restore again (platform-specific solution case → `NoRestore` stays false), increasing the chance this fires.
- **Show me:** Configure a source like `https://feed.example/v3/index.json?sig=SECRET`, run `winapp run App.slnx --quiet` where the build performs the restore → a NuGet feed-index error prints `...?sig=SECRET` to stderr. Expected (matching the restore path): `...?<redacted>`.
- **Why it matters:** Authenticated feed URLs can land in terminal scrollback and CI logs — the exact leak this PR exists to prevent.
- **Smallest fix:** Wrap the `--json`/`--quiet` build output callbacks in `NugetErrorMessage.Redact` (one line, mirroring the restore path). The native-terminal inherited-stdio case can't be redacted without dropping dotnet's terminal logger — document that as a known limitation, or avoid the implicit build restore there.
- **Location:** `src/winapp-CLI/WinApp.Cli/Services/ProjectRunService.cs:710-713` (and inherited path at `:732`)

**2. User-supplied `-p Platform=x64` on a solution skips the build restore it needs**
- **What is wrong:** The new `NoRestore` decision only inspects `options.Platform`. When the user passes `-p Platform=x64`, `ResolvePlatformInjection` intentionally leaves `options.Platform` **null** (proven by `ResolvePlatformInjection_UserSuppliedPlatform_IsRespected`) and forwards the value via `options.Properties`. So the whole-solution restore drops the Platform (correct, avoids MSB4126), `NoRestore` is set to true because `options.Platform` is null — yet the build still passes `-p:Platform=x64`. The platform-conditioned assets are never restored for the target.
- **Show me:** all-managed `.slnx` + `winapp run App.slnx --project App -p Platform=x64` → `dotnet restore App.slnx` runs without Platform, then `dotnet build App.csproj --no-restore -p:Platform=x64`. Expected: the target restores again (as it correctly does for a *resolved* Platform).
- **Why it matters:** The PR's own fix for platform-conditioned assets covers the resolved-Platform case but leaves the user-supplied-Platform case broken — a user with platform-split `PackageReference`s hits missing assets. Before this PR the MSB4126 failure fell back to a per-project restore that carried the Platform, so this is a regression.
- **Smallest fix:** In the `NoRestore` decision, treat a user-supplied `Platform=` in `options.Properties` as an effective platform (keep `NoRestore=false` after the solution restore), not just `options.Platform`.
- **Location:** `src/winapp-CLI/WinApp.Cli/Services/ProjectRunService.cs:180-188`

### Non-blocking

**Sample README says directory mode finds the `.sln`** — The sample now ships both `WinUISolution.sln` and `WinUISolution.slnx`; `CollapseSolutionMigrationPairs` collapses the pair preferring `.slnx`, so `winapp run .` uses the `.slnx`, but the README still says it finds the `.sln`. Fix: say directory mode finds the solution (the `.slnx` when both are present). `samples/winui-solution/README.md:68`

**Docs promise the "exact" invocation, but it's redacted** — `docs/usage.md` says winapp prints each *exact* `dotnet restore`/`dotnet build` invocation, but the display path redacts secret-like properties and signed feed URLs first. Fix: change "exact" to "sanitized" / "redacted when needed." `docs/usage.md:790`

### What was exercised
- `dotnet build WinApp.Cli.csproj -c Debug` — succeeded, 0 warnings/errors.
- `WinApp.Cli.Tests.exe --filter FullyQualifiedName~ProjectRunServiceTests` — 203/203 passed, confirming the branch's tests are green and the two findings are genuine gaps, not broken-build artifacts.
- Not exercised at runtime: an actual credentialed-feed leak and an actual platform-split missing-asset build — both require network access to a private feed (unreachable from this machine) and specific project setups. Both were instead confirmed by tracing the code paths end-to-end plus the existing `ResolvePlatformInjection_UserSuppliedPlatform_IsRespected` test that proves `options.Platform` is null for user-supplied Platform.

<sub>🤖 Generated with multi-dimensional pr-review (security, correctness, cli-ux, alternative-solution, ship-surfaces, necessity, + different-model cross-check).</sub>

zateutsch (commented):
Findings attached above.

zateutsch:
## PR Re-Review (round 2)

The branch advanced since the last review (`4b493998 Merge latest main and address review feedback`). This is a re-review, so I focused on critical/high and treated the fix code as re-openable.

### Prior findings — status

| Prior finding | Status |
|---|---|
| 🔴 Build-triggered restore leaks feed credentials | ✅ **Fixed & verified.** `--json`/`--quiet` build output now goes through `NugetErrorMessage.Redact`, and `nativeTerminal &= options.NoRestore` forces the streamed+redacted path whenever the build might restore. |
| 🟠 User `-p Platform=x64` skips build restore | ⚠️ **Partially fixed** — see below. New `HasEffectivePlatform` fixes the plain form; a packed-property edge remains. |
| Doc: "exact" invocation | ✅ Fixed — `docs/usage.md:790` now says "sanitized". |
| Doc: README directory mode | ✅ Fixed — now "directory mode prefers the .slnx migration pair". |

### Remaining finding

**🟠 HIGH — Packed `Platform` property still reintroduces MSB4126 on a configuration-free `.slnx`**
- **What is wrong:** The fix made *detection* segment-aware (`HasEffectivePlatform` → `UserSpecifiesProperty` → `PropertySegments`, which splits on `;` and `,`), but the solution Platform-*strip* is still whole-string: `property.StartsWith("Platform=", …)`. A packed `-p` value whose Platform isn't the first segment is never stripped from the solution restore.
- **Show me:** `winapp run WinUISolution.slnx -p Flavor=Retail;Platform=x64` → `BuildRestorePassArguments(solution, …)` emits `-p:Flavor=Retail;Platform=x64` → MSBuild reads solution `Platform=x64` → **MSB4126** on the configuration-free `.slnx`. (Conversely `-p Platform=x64;Flavor=Retail` starts with `Platform=`, so the *whole* property is dropped, silently losing `Flavor=Retail` from the restore.) Expected: strip only the `Platform` segment, keep the rest.
- **Why it matters:** This is the exact error the PR exists to eliminate, and the repo already treats packed `-p` as a first-class case it defends consistently everywhere else (`IsDedicatedFlagProperty`). A user hits the failure the PR advertises as fixed — false confidence.
- **Smallest fix:** Make the strip segment-aware, reusing the existing `PropertySegments`/`PropertyName` helpers — drop only the `Platform` segment and re-join the survivors, instead of `StartsWith`. Add tests for `Flavor=Retail;Platform=x64` and `Platform=x64;Flavor=Retail`.
- **Location:** `src/winapp-CLI/WinApp.Cli/Services/ProjectRunService.Arguments.cs:63-66`

### Validation
- `dotnet build WinApp.Cli.csproj -c Debug` → succeeded, 0 warnings/errors.
- `WinApp.Cli.Tests.exe --filter FullyQualifiedName~ProjectRunService` → **203/203 passed**, confirming the finding is an uncovered edge, not a broken build. No existing test packs `Platform` with another segment.
- Not runtime-reproduced: MSB4126 needs a real `.slnx` + MSBuild + a packed non-dedicated property. Confirmed instead by direct code trace — the decisive fact (`"Flavor=Retail;Platform=x64".StartsWith("Platform=")` is false → not stripped) is static and unambiguous.

Everything else (cli-ux, alternatives, necessity, ship-surfaces) was clean or resolved last round and is unchanged.

<sub>🤖 Generated with multi-dimensional pr-review (security + correctness re-review + code-path validation).</sub>

nmetulev:
@zateutsch Thanks for the re-review. I validated the packed-property case against the published CLI rather than the internal argument builder. Project mode deliberately rejects a single `-p` containing `;` or `,` before any restore/build starts:

`winapp run WinUISolution.slnx --json -p "Flavor=Retail;Platform=x64"` -> exits nonzero with `Invalid --property 'Flavor'`; no `dotnet` process is invoked and `MSB4126` is not reachable. This is covered by `ProjectMode_SemicolonPackedProperty_IsRejectedBeforeBuilding` (with equivalent comma coverage), so I did not add segment-aware stripping to a boundary that validated CLI input cannot reach.

The independent re-review did find one separate credential-redaction gap: malformed authenticated URLs could fail URI parsing and be returned unchanged. Commit `fb0aeda4a` now fails closed to `<redacted-url>` when an unparseable URL-shaped value contains query or user-info markers. The published-CLI reproduction now reports `SECRET_COUNT=0` and two redacted command occurrences; the focused restore/redaction suite passes 244/244.

src/winapp-CLI/WinApp.Cli.Tests/ProjectRunServiceTests.cs
```diff
@@ -3451,6 +3864,38 @@ public async Task RunBuildPassAsync_RealTerminal_UsesInheritedLauncherWithoutTlO
...
+        var csproj = WriteFile("App.csproj", ExecutableCsproj);
+        var dotnet = new FakeDotNetService
+        {
+            RunDotnetStreamingHandler = (_, onOut, _) =>
+            {
+                onOut?.Invoke("NU1301 https://feed.example/v3/index.json?sig=BUILD_RESTORE_SECRET");
+                return 0;
+            },
+        };
+        var console = new TestConsole();
```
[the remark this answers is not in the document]
nmetulev: Fixed in `4c4cad52d`. The test now declares its local `TestConsole` with `using var`, so it is disposed on every exit path without changing the test behavior.

zateutsch:
## PR Re-Review (round 3)

Reviewed at `4c4cad52`. This is a re-review, so critical/high only. Build green; `WinApp.Cli.Tests.exe` (ProjectRunService + NugetErrorMessage) = **271/271 passed**.

### Resolved since last round ✅
- 🔴 **Build-triggered restore credential leak** — fixed. The `--json`/`--quiet` build output now runs through `NugetErrorMessage.Redact`, and `nativeTerminal &= options.NoRestore` keeps the unredacted inherited-stdio path off whenever the build might restore.
- 🟠 **Plain `-p Platform=x64` skipping the build restore** — fixed via the new segment-aware `HasEffectivePlatform` detection.
- **`NugetErrorMessage` fail-closed hardening** (`fb0aeda4`) — reviewed, no findings. Unparseable URL-shaped values containing `?`/`@` now redact to `<redacted-url>`. Well covered by the new tests.
- Both doc items ("exact"→"sanitized"; README directory-mode wording) — fixed.

### Still open

**🟠 HIGH — Packed `Platform` property still reintroduces MSB4126 on a configuration-free `.slnx`**
- **What is wrong:** The fix made *detection* segment-aware (`HasEffectivePlatform` → `UserSpecifiesProperty` → `PropertySegments`, splitting on `;`/`,`), but the solution Platform-*strip* is still whole-string `property.StartsWith("Platform=", …)`. Detection and stripping are now inconsistent.
- **Show me:** `winapp run WinUISolution.slnx -p Flavor=Retail;Platform=x64` → `BuildRestorePassArguments(solution, …)` emits `-p:Flavor=Retail;Platform=x64` → MSBuild reads solution `Platform=x64` → **MSB4126** on the configuration-free `.slnx`. The reversed form `Platform=x64;Flavor=Retail` starts with `Platform=`, so the *whole* property is dropped instead — silently losing `Flavor=Retail` from restore.
- **Why it matters:** When it hits, it produces exactly the MSB4126 error this PR's feature is meant to eliminate — reads as a regression against the PR's own promise. The rest of the file already treats packed `-p` as first-class (`IsDedicatedFlagProperty`), so this is the one inconsistent site.
- **Likelihood / severity note:** trigger conditions are narrow (a *packed* `-p` bundling `Platform` with another property, AND a configuration-free `.slnx` specifically). Separate `-p` flags and classic `.sln` are unaffected. Reasonable to treat as a fast-follow rather than a hard merge-blocker — but it is a cheap fix.
- **Smallest fix:** strip only the `Platform` segment using the existing `PropertySegments`/`PropertyName` helpers and re-join survivors, instead of `StartsWith`. Add tests for `Flavor=Retail;Platform=x64` and `Platform=x64;Flavor=Retail`.
- **Location:** `src/winapp-CLI/WinApp.Cli/Services/ProjectRunService.Arguments.cs:63-66`

Everything else (cli-ux, alternatives, necessity, ship-surfaces) was clean or resolved earlier and is unchanged.

<sub>🤖 Generated with multi-dimensional pr-review (security + correctness re-review + code-path validation).</sub>

zateutsch (changes requested):
One hanging problem from last review. Only finding, should be good to go once fixed.

nmetulev:
@zateutsch I double-checked the remaining packed-`Platform` finding against the current source, focused tests, and a freshly published NativeAOT CLI. It is not reachable through `winapp run`.

`RunCommand.ProjectMode` calls the shared `MsBuildPropertyValidator.Validate` before constructing `ProjectRunOptions` or calling `IProjectRunService.BuildAndResolveAsync`. That validator rejects any single `-p` containing `;` or `,`. The internal `BuildRestorePassArguments` method assumes this validated model; calling it directly bypasses the public command boundary.

Fresh published-binary results:

```text
-p "Flavor=Retail;Platform=x64"
  exit code: 1
  error: Invalid --property 'Flavor'. A single -p cannot pack multiple properties...
  dotnet invocations: 0
  MSB4126 occurrences: 0

-p "Platform=x64;Flavor=Retail"
  exit code: 1
  error: Invalid --property 'Platform'. A single -p cannot pack multiple properties...
  dotnet invocations: 0
  MSB4126 occurrences: 0
```

The supported syntax behaves correctly:

```text
-p Flavor=Retail -p Platform=ARM64
  exit code: 0
  solution restore: keeps Flavor=Retail, omits Platform
  project build: keeps Flavor=Retail and Platform=ARM64
  MSB4126 occurrences: 0
```

The relevant validation suite also passes 21/21, including `ProjectMode_SemicolonPackedProperty_IsRejectedBeforeBuilding`, comma-packed coverage, and the shared `MsBuildPropertyValidatorTests`.

Because both examples in the finding are rejected before the cited code path, segment-aware stripping would add handling for an invalid state rather than fix a user-reachable regression. Could you please dismiss the changes-requested review or approve the current revision?

zateutsch (approved):

PR #791, opened 2026-08-26
Polish project restore output modes

nmetulev:
## Summary
- pass quiet verbosity to project-mode `dotnet restore` while preserving default, JSON, and interactive terminal behavior
- remove duplicate restore command text from verbose decision traces while keeping displayed commands redacted
- add direct JSON/quiet restore routing regressions and update project-mode docs and setup guidance

## Validation
- `dotnet test src\winapp-CLI\WinApp.Cli.Tests\WinApp.Cli.Tests.csproj -c Release --filter "FullyQualifiedName~ProjectRunServiceTests"` (195 passed)
- Release CLI against a throwaway two-project solution: quiet emitted no restore invocation or normal restore progress; verbose emitted one restore invocation
- `.\scripts\validate-plugin-package.ps1`
- `git diff --check`
- `.\scripts\build-cli.ps1` reached NativeAOT publish, then stopped at the machine's known missing Desktop Development for C++ linker prerequisite

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -754,7 +754,7 @@ Project mode requires the **.NET SDK 8.0.100 or newer** (for MSBuild `--getPrope
 - `--no-restore` - Skip restoring the project before building.
 - `-p, --property <Name=Value>` - MSBuild property, forwarded to both the build and the property evaluation. Repeatable (e.g. `-p WindowsPackageType=None`).
 
-**Build output & verbosity:** dependency restore and `dotnet build` output **stream live** to your console, followed by a fast property-evaluation pass. In an interactive terminal, dotnet's terminal logger shows in-place progress and elapsed time. For redirected output and CI, winapp streams plain lines instead. It prints each exact `dotnet restore …` or `dotnet build …` invocation first, so package downloads, feed retries, errors, and build warnings remain visible. Verbosity:
+**Build output & verbosity:** dependency restore and `dotnet build` output **stream live** to your console, followed by a fast property-evaluation pass. In an interactive terminal, dotnet's terminal logger shows in-place progress and elapsed time. For redirected output and CI, winapp streams plain lines instead. In default and verbose modes, winapp prints each exact `dotnet restore …` or `dotnet build …` invocation first, so package downloads, feed retries, errors, and build warnings remain visible. Verbosity:
 
 | Flag | dotnet verbosity | Adds |
 |------|------------------|------|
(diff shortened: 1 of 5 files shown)


azchohfi:

# PR Review

## Decision

**Merge.** No critical or high issues. I confirmed all four output modes behave as the updated docs describe by running the published CLI against a real WinUI app. One optional test-quality nit below.

## Must fix

None — mergeable as-is.

## Non-blocking

### The new `--json` restore test asserts on a value the test built itself

- **What is wrong:** The test builds restore arguments locally without passing a verbosity, then asserts those same arguments lack `-v quiet`. The assertion cannot fail, so it does not test production behavior.
- **Show me:** Line 2946 calls `BuildRestorePassArguments(solution, options)` (no verbosity argument); line 2979 asserts that local `args` does not contain `-v quiet`. If `ResolveRestoreVerbosity` regressed to return `"quiet"` for JSON, the three production call sites would pass `-v quiet` — and this test would still pass.
- **Why it matters:** This PR's purpose is pinning these output modes; a vacuous assertion leaves the `--json` half unpinned while reading as covered.
- **Smallest fix:** Assert on the invocation recorded by a `BuildAndResolveAsync` run with `Json: true`, the way the sibling quiet test already does via `dotnet.StreamingCalls`.
- **Location:** `src/winapp-CLI/WinApp.Cli.Tests/ProjectRunServiceTests.cs:2946-2980`

## What was exercised

Harness: a real WinUI app (`microsoft-ui-reactor`'s ReactorGallery — `.slnx` plus 8 project references), driven by a published Release `winapp.exe` rather than `dotnet run`.

- `dotnet publish -c Release` — succeeded.
- `dotnet test --filter ProjectRunServiceTests` — 195 passed, 0 failed.
- `winapp run` (default) — stdout carries the banner and the exact `dotnet restore …` invocation, then live restore output. Matches the doc claim.
- `winapp run --quiet` — **stdout 0 bytes**; restore/build output on stderr; zero occurrences of the invocation *or* the "Restoring … dependencies…" banner. Suppression confirmed.
- `winapp run --json` — **stdout is 26 bytes of valid JSON** (`{"ProcessId":…}`, parsed successfully); the invocation and ~9 KB of child output went to stderr. Confirms `--json` keeps default restore verbosity rather than quiet.
- `winapp run --verbose` — still prints the exact `dotnet restore …` invocation plus the new decision trace. This addresses the main risk in the diff: replacing the `LogDebug` that carried the redacted arguments costs `--verbose` users nothing, because `RunRestoreCommandAsync` already echoes the invocation.
- App built and launched end-to-end; processes cleaned up afterward.

**Not exercised:** the interactive-terminal path (dotnet's in-place terminal logger) — all runs redirected stdout/stderr, so only the plain-line/CI path was validated. Two of the three changed call sites (`RunRestorePassAsync` for SDK-less CsWinRT resolution, and the solution-sibling loop) were not hit at runtime; the gallery took the all-managed solution restore path. Both are covered by the passing unit tests and use the identical one-line expression.

zateutsch (approved):
Review was fully clean for me.

nmetulev:
Superseded by #789, which now includes the output-mode polish and all follow-up review fixes.

Issue #703, opened 2026-07-30
[Feature]: Publish sample indexes upstream (WinUI Gallery + Community Toolkit) so find-ui stops scraping


nmetulev:
### Is your feature request related to a problem? Please describe.

`winapp find-ui` (#681) builds its corpus by scraping three upstream repos. The cost of that scraping is wildly uneven, and it comes down entirely to **what each repo publishes**:

| Source | Publishes | Fetcher size |
|---|---|---|
| `microsoft/microsoft-ui-reactor` | [`reactor-search-index.json`](https://github.com/microsoft/microsoft-ui-reactor/blob/main/samples/ReactorGallery/reactor-search-index.json) — metadata **and sample code** | **7 KB** (`ReactorFetcher.cs`) |
| `microsoft/WinUI-Gallery` | `ControlInfoData.json` — metadata only | **43 KB** (`GalleryFetcher.cs`) |
| `CommunityToolkit/Windows` | nothing published | **54 KB** (`ToolkitFetcher.cs`) |

Reactor's fetcher is ~6x smaller for the same job, purely because the Reactor team publishes a purpose-built index. Gallery and Toolkit require us to reverse-engineer repo layout, split samples out of XAML pages, clean code-behind, fold `#if` branches, re-balance tags, and truncate — roughly **1,450 lines of parsing** that exists only to reconstruct data the upstream repos already have.

This is not just a size problem. It produces **wrong output**:

- **9 of 321 Gallery scenarios (~3%) emit structurally malformed XAML** that fails to parse. Verified against both the current corpus and the original `winui-search` snapshot in `microsoft/win-dev-skills` — identical results, so this is inherent to scraping, not a porting bug:
  ```
  xamlresources-3, appbarbutton-5, swipecontrol-4, swipecontrol-5, listview-8,
  infobadge-2, selectorbar-3, themeshadow-1, richtextblock-3
  ```
  Example — `gallery-appbarbutton-5`, a short, **non-truncated** sample:
  ```xml
  <AppBarButton Icon="Save" Label="Save" Click="AppBarButton_Click">
    <AppBarButton.KeyboardAccelerators>
      <KeyboardAccelerator Modifiers="Control" Key="S"/>
    <AppBarButton.KeyboardAccelerators/>
  </AppBarButton></AppBarButton.KeyboardAccelerators></AppBarButton>
  ```
[...]

PR #806, opened 2026-09-02
feat: define the shared WinUI sample index contract (#703 Phase 0)

### LINKED ISSUE #808: Phase 0: define the shared WinUI sample index contract
Parent: #703

Define the shared sample-index contract **inside winappCli**, before asking either upstream repo for anything. We are the consumer, so we define the format — each upstream ask then arrives as "here is the schema, here is the file it produces, here is the tested code that produced it" rather than a request to design a format for us.

## Scope

- [x] A published JSON Schema for the index — `docs/winui-sample-index.schema.json`
- [x] A generic reader/writer pair, with `ReactorFetcher` rewired onto them as proof the shared reader works against a real published index
- [x] A reference generator that emits an index for all three sources on demand
- [x] A validation harness proving the schema carries the real corpus losslessly

## Why the vocabulary is Reactor's

`microsoft/microsoft-ui-reactor` already publishes `reactor-search-index.json`. This contract generalizes that rather than competing with it, so Reactor's live index validates against the schema today with no change on their side.

## Not in scope

No upstream contact, no user-facing behavior change, and nothing committed for the generated indexes.

Jaylyn-Barbee:
## Description

[#703](https://github.com/microsoft/winappCli/issues/703) asks WinUI-Gallery and CommunityToolkit to publish machine-readable sample indexes so `find-ui` stops scraping their repository layouts. Neither team has ever been asked for this — research found **zero** prior art in either repo — so appetite is genuinely unknown, and that is the main risk on the issue.

This is **Phase 0**, which is entirely on our side of the fence. We are the consumer, so we define the contract first. The point is that each upstream ask arrives as *"here is the schema, here is the file it produces, here is the tested code that produced it"* rather than asking two maintainer teams to design a format for us.

Nothing here changes user-facing behavior.

**What ships:**

| Piece | What it is |
|---|---|
| `docs/winui-sample-index.schema.json` | The published contract. JSON Schema 2020-12, `schemaVersion` pinned to `const: 1`. |
| `SampleIndexParser` | Generic index → `Scenario[]`. Kind-checks every element, since this parses untrusted network content. |
| `SampleIndexWriter` | Reference generator core. Deterministic output. Lives in the **build-only baker**, not the shipped CLI. |
| `ReactorFetcher` | Rewired onto the shared parser: **151 → 67 lines**, no parsing logic of its own. |
| `--emit-index` | Opt-in flag on the snapshot baker that writes an index per source. Never committed, never shipped. |

**The vocabulary is deliberately Reactor's.** `microsoft-ui-reactor` already publishes `reactor-search-index.json`, and this contract generalizes it rather than competing with it. Field names are the ones Reactor already ships, which means **Reactor's live index validates against this schema today with no change on their side** — that is the proof the contract is adoptable rather than aspirational, and `ReactorFetcher` reading it through the shared parser is the working demonstration.

## Usage Example

Not a shipped CLI surface. The generator is opt-in on the build-only baker:

```bash
# Generate an index per source into a scratch directory (never committed)
dotnet run --project src/winapp-CLI/WinApp.Cli.SnapshotBaker -- --emit-index ./artifacts/indexes
```

Verified live against all three sources. This is a **usable-only** export: scenarios are sanitized first and any left with neither XAML nor C# are omitted, so gallery writes 305 of the 323 it fetches. Those 18 are already dropped before search today, so including them would hand upstream known-broken content to publish. "Lossless" here is a property of the schema — a scenario that survives is carried field for field — not a promise that every fetched scenario appears.

| Source | Controls / samples | Size |
|---|---|---|
| gallery | 106 / 305 | 375 KB |
| toolkit | 26 / 48 | 98 KB |
| reactor | 95 / 95 | 83 KB |

## Related Issue

Closes #808 — Phase 0 of #703.

## Type of Change

- ✨ New feature

## Checklist

- [x] New tests added for new functionality — 18 in `SampleIndexTests.cs`
- [ ] Main [README.md](../README.md) updated — n/a, no user-facing surface
- [ ] [docs/usage.md](../docs/usage.md) updated — n/a, no CLI command changed; the baker is a build-only tool and, like the existing bake path, documents itself through its own usage output
- [ ] [Language-specific guides](../docs/guides) updated — n/a
- [ ] [Sample projects updated](../samples) — n/a
- [ ] Shipped skills updated in `plugins/winapp/skills/` — n/a, no CLI workflow changed

### This PR needs the `corpus-rebake-not-needed` label, and here is the evidence

`find-ui-corpus-check.yml` fires on two changed paths — `WinApp.Cli.SnapshotBaker/` and `ReactorFetcher.cs` — and this PR re-bakes nothing. That is correct behavior from the check and the right question to ask, so it was answered directly rather than waved off:

> The full corpus was baked from `origin/main` and from this branch back to back, and the decompressed snapshots compared.

```
gallery  main=62016BF907F0  branch=62016BF907F0  720,910 chars  IDENTICAL
toolkit  main=53331F9368D4  branch=53331F9368D4  122,110 chars  IDENTICAL
reactor  main=2D1815A46E13  branch=2D1815A46E13   97,394 chars  IDENTICAL
```

Rewiring `ReactorFetcher` onto the shared parser (−97/+20 lines) is **behavior-preserving**. Gallery and Toolkit are the control: their fetchers are untouched and also came out identical, which confirms both bakes saw the same upstream state, so a reactor difference would have been attributable to this code rather than to upstream moving mid-test.

Separately and unrelated to this PR: the committed corpus is now behind upstream (gallery 321 → 323, reactor 93 → 95, baked 2026-08-11). That is exactly what `find-ui-corpus-drift.yml` exists to report, and it is deliberately not folded in here.

### Two schema defects that only showed up against real data

Both were found by round-tripping all three **live** corpora, and both would have shipped as silent quality regressions once a source published an index.

**1. Curated keywords are a separate ranking signal.** `ProviderData` carries two dictionaries, not one: `SearchEngine` weights `Keywords` at **5.0** and `Tags` at **3.0**. Only the Toolkit populates `Keywords`, from author-written markdown frontmatter. A single `keywords` field would have folded the Toolkit's best signal down to 3.0 and quietly re-ranked results.

`keywords` keeps its published meaning (`Tags`, 3.0) because Reactor already ships that field and re-pointing it would change live rankings for no reason. The new `curatedKeywords` carries the 5.0 slot. Against the live Toolkit corpus, **26 of 26** controls populate both — so this is load-bearing, not hypothetical.

**2. `details` and `xmlnsImports` belong to the sample, not only the control.** The writer took control metadata from the first scenario in each group. For `description`, `apiNamespace` and friends that is correct — they are control facts. For these two it is not: each Toolkit sample is its own markdown file with its own XAML, so siblings legitimately disagree. `segmented` has five samples with five descriptions and differing import counts (2/2/2/2/**3**). Collapsing them onto the first would have emitted XAML using a prefix the index never declared.

Both fields are now valid at either level. The writer **hoists them to the control when every sample agrees** and writes them per sample when they do not, so the common case stays compact — Gallery, where all samples share one description, grows 412 bytes; Toolkit grows 5.7 KB; Reactor not at all. The reader prefers the sample value and falls back to the control's.

### Two findings about existing data, recorded rather than fixed

- **`gallery/itemsrepeater` has one scenario scraped with no control metadata at all** — `ApiNamespace`, `ControlDescription`, `Description`, `RelatedControls` and `Docs` are all empty on `itemsrepeater-7` while its six siblings have them. Round-tripping *fills* it from the siblings. That is a repair, not a loss, so the contract asserted in tests is **"a populated value is never changed or dropped"** rather than strict equality. Fixing the scrape is out of scope here and moot once Gallery publishes an index.
- **`gallery-tags.json` holds 13 entries for control ids that have no scenarios** (`scratchpad`, `listbox`, `wrappanel`, `jumplist-gallery`, …). `SearchEngine` derives its control list from scenarios, so these are already dead today. The index cannot carry them and does not need to.

### Design calls worth reviewing

- **`additionalProperties: true` throughout.** Reactor's live index carries `generatedFrom`, `category` and `galleryRoute`, which this contract does not model. A closed schema would reject the one real-world index that already exists, which would be an odd contract to hand someone.
- **`details` is a new field rather than a redefinition of `description`.** Reactor's published `description` already means the one-line summary. Gallery keeps both a short subtitle and a long-form description, so it needs a second field — taking `description` for the long form would have broken the one index in the wild.
- **An absent `schemaVersion` is accepted as v1; an unknown one is refused.** A future v2 index should fail loudly rather than be half-parsed into wrong results.
- **Determinism is enforced, not incidental.** Controls are sorted by id, `generatedAtUtc` is omitted by default, and `JavaScriptEncoder.UnsafeRelaxedJsonEscaping` keeps XAML `<` from becoming `\u003C`. A generated artifact that churns between runs cannot be diffed by a drift check and is unreviewable by the maintainers we are asking to host it.
- **The generator writes on demand and commits nothing.** This follows the precedent set when #770 merged with only the compressed blobs checked in. The existing positional bake invocation used by `build-cli.ps1` and the drift workflow is deliberately byte-identical.

### Review feedback

**The writer moved out of the shipped CLI.** `--emit-index` was always a baker flag, but `SampleIndexWriter` itself sat in `WinApp.Cli/Services/Controls/`, so the shipped binary carried code whose only caller is a build-time tool. It now lives in `WinApp.Cli.SnapshotBaker/`. The CLI keeps only `SampleIndexSchema` and `SampleIndexParser` — the parts that survive into a world where upstream publishes and we only ever *read*. Moving it also took the file out of the `Services/Controls/` `.editorconfig` subtree, so it now meets the repo's brace and file-header rules.

**A malformed `language` tag no longer yields C#.** `GetString` collapsed *absent* and *present-but-not-a-string* into `""`, so `"language": 42` parsed as C# and would have been offered as paste-ready code. `IsCSharp` now inspects the `JsonElement` directly and treats a non-string tag as not-C#. Covered for `42`, `null` and `""`.

**Zach's blocking finding is fixed in #815, not here.** He flagged that `--emit-index` emits winapp-authored samples presented as Gallery content, and proposed filtering them at emit time. Investigating it showed the divergence was not an emit-time problem: `GalleryFetcher` was injecting and rewriting samples in the *fetch* path and stamping them `Source = "gallery"`, so they were already reaching users mislabeled, long before any index was written. Filtering at emit would have hidden the symptom on the one surface we were about to show upstream. #815 removes the divergence at the source instead, which also makes the filter unnecessary. This PR is therefore unchanged in that respect and its byte-identical-corpus evidence below still stands.
### Verification

- **209/209** targeted tests pass (`SampleIndex`, `ReactorFetcher`, `ScenarioSanitizer`, `Controls`, `FindUi`, `Gallery`, `Toolkit`, `Search`, `EmbeddedSnapshot`).
- Release build clean — **0 warnings**, warnings-as-errors on.
- A throwaway harness fetched all three sources live and round-tripped every scenario through writer and reader, asserting scenario-level equality, no lost live tag, determinism against real upstream data, and identical sanitizer behavior before and after. It is deleted; its durable conclusions are three hermetic tests covering per-sample divergence, control-level hoisting, and reader inheritance.
- `Schema_DeclaresExactlyTheFieldsTheReaderAndWriterUse` asserts `docs/winui-sample-index.schema.json` and `SampleIndexSchema` cannot drift apart — the schema file is copied to the test output for that check.


zateutsch (commented):

## Decision
**Changes required** — the generated Gallery index includes a winapp-authored sample and presents it as upstream Gallery content. The parser’s handling of malformed `language` values is also worth tightening, but it is non-blocking.

## Must fix
### Generated Gallery indexes misattribute locally authored samples
- **What is wrong:** Index generation writes the post-processed Gallery corpus, including scenarios injected locally by winapp rather than supplied by WinUI Gallery.
- **Show me:** Running `WinApp.Cli.SnapshotBaker.exe --emit-index <dir>` produced `index-gallery.json` containing `Photo gallery: image grid (UniformGridLayout)`, which is added by `GalleryFetcher.InjectMissing`; expected: an upstream index containing only Gallery-owned samples.
- **Why it matters:** The artifact intended for WinUI Gallery maintainers would ask them to publish code their repository does not own, making the proposed authoritative index inaccurate.
- **Smallest fix:** Mark locally injected scenarios as synthetic and exclude them in `EmitIndexesAsync`, with a regression test proving emitted source indexes contain fetched source content only.
- **Location:** `src/winapp-CLI/WinApp.Cli.SnapshotBaker/SnapshotBaker.cs:211`

## Non-blocking
### Malformed language values are accepted as C#
- **What is wrong:** A missing `language` and a present non-string value both become an empty string, which the parser treats as C#.
- **Show me:** `{"code":"auto x = 1;","language":42}` flows through `GetString` and is retained as paste-ready C#; expected: reject the code because a present language must be the string `csharp`.
- **Why it matters:** Malformed upstream data can surface a wrong-language snippet that fails when pasted into a C# project. Current published Reactor data uses valid `csharp` values, so this is not blocking.
- **Smallest fix:** Distinguish an absent property from a present non-string property and add that malformed case to parser tests.
- **Location:** `src/winapp-CLI/WinApp.Cli/Services/Controls/SampleIndexParser.cs:121`

## What was exercised
- `dotnet build src\winapp-CLI\WinApp.Cli.SnapshotBaker\WinApp.Cli.SnapshotBaker.csproj -c Debug` — succeeded with 0 warnings and 0 errors.
- Filtered `SampleIndexTests` — all 21 passed.
- Built `WinApp.Cli.SnapshotBaker.exe --emit-index <scratch-dir>` — fetched all three providers and reproduced the Gallery provenance defect in generated output.
- `scripts\build-cli.ps1` — compilation and generation progressed into the full test suite, then repeatedly failed downloading `Microsoft.Windows.SDK.BuildTools` and stopped producing output; the stalled run was stopped. Internal-feed authentication was unavailable in this shell, so the full suite did not complete.

zateutsch:

## Decision
**Merge.** This is a well-scoped Phase 0 for #703: a published sample-index schema, a shipped reader that Reactor already uses, and a build-only writer + `--emit-index` flag whose only job is to prove the schema round-trips. Build is clean, 22 tests pass, and no security, correctness, or ship-surface issues survived review. One minor, optional UX nit in an internal tool.

## Must fix
None — mergeable as-is.

## Non-blocking
### Baker treats an unknown flag as the output directory
- **What is wrong:** In the build-only `SnapshotBaker`, any positional arg that isn't `--emit-index` becomes the output directory, including a mistyped flag.
- **Show me:** `dotnet run --project src/winapp-CLI/WinApp.Cli.SnapshotBaker -- --bogus` → starts the default network bake into a directory literally named `--bogus` → expected: usage error, exit 1.
- **Why it matters:** A typo silently kicks off network/file work in the wrong place instead of failing fast. Low impact since this tool is maintainer-only, not shipped.
- **Smallest fix:** Reject a positional value starting with `-` unless it is a supported flag, before assigning `outputDirectory` (still allows intentional paths like `.\-out`).
- **Location:** `src/winapp-CLI/WinApp.Cli.SnapshotBaker/Program.cs:27-43`

## What was exercised
- `dotnet build WinApp.Cli.SnapshotBaker.csproj -c Debug` (pulls in WinApp.Cli) — succeeded, 0 warnings.
- `WinApp.Cli.Tests.exe --filter FullyQualifiedName~SampleIndexTests` — 22/22 passed (round-trip fidelity, mixed per-sample vs hoisted `details`/`xmlnsImports`, non-C# language drop, oversized-usings guard, unknown schema-version refusal, non-object tolerance).
- Reactor-parity regression check (old inline parser vs new `SampleIndexParser` delegation) — no scenario/tag dropped: Reactor's index has no `schemaVersion` (accepted as v1) and no `language` tags (accepted as C#).
- Untrusted-input trace: parsed index values reach display-only markdown; `ScenarioSanitizer` strips terminal escapes on both read (`ControlsSearchService`) and write (`EmitIndexesAsync`) paths; no value reaches `Process.Start` or a file-path sink; `docs.uri` is never emitted.
- Not exercised: a live `--emit-index` network run and the non-blocking finding's failure path — both confirmed by static read of `Program.cs`.

zateutsch (approved):
Approved with one non blocking comment.

PR #815, opened 2026-09-03
fix(find-ui): Gallery corpus is upstream-only; ten controls that returned no code now do

Jaylyn-Barbee:
## Description

Two things, one cause.

**1. `GalleryFetcher` presented winapp-authored content as WinUI Gallery's.** It post-processed the corpus it fetched, then stamped the results `Source = "gallery"` — the value `find-ui` renders as the `[gallery]` tag and uses to build ids like `gallery-itemsrepeater-7`.

| Mechanism | What it did | State when found |
|---|---|---|
| `InjectMissing` — CommandBar | Appended a hand-written CommandBar sample | **Dead.** Guarded on Gallery having no commandbar sample; Gallery ships one now. |
| `InjectMissing` — `itemsrepeater-7` | Appended a hand-written `UniformGridLayout` photo grid | **Live, unconditional.** No detection logic at all. |
| `ApplyOverrides` — `tabview-1` | Rewrote upstream's `CreateNewTab` C# body | **Live.** Guarded on `CSharp.Contains("Frame")`. |

All three are gone. `FetchAsync` is now a passthrough.

This blocks [#703](https://github.com/microsoft/winappCli/issues/703): we cannot ask WinUI-Gallery to publish a sample index while merging samples they do not own into their data — still less hand them a *generator* that does it.

**2. The CommandBar injection was papering over a bug winapp owned.** Review asked why CommandBar needed a hand-written sample at all. It does not. Upstream's sample is complete and well-formed; winapp's extractor was destroying it.

Gallery samples declare optional attributes and injected markup as `$(Name)` tokens that the Gallery app expands at runtime. `GalleryFetcher` flattened **every** token to `"..."` regardless of position, which is only valid in one of the three:

```
upstream   <Button Content="Standard XAML button" Click="Button_Click" $(IsEnabled)/>
flattened  <Button Content="Standard XAML button" Click="Button_Click" .../>
           -> Name cannot begin with the '.' character
```

`ScenarioSanitizer.XamlIsWellFormed` then rejected the snippet and `Xaml` was set to `null`. The control still appeared with a fetchable id and a "Fetch full code" prompt, and returned **a successful-looking result with no code in it**.

Surveyed all 480 upstream sample files. **Ten were affected:**

`Button` · `ToggleButton` · `RepeatButton` · `HyperlinkButton` · `ProgressRing` (×2) · `CommandBar` · `AnimatedIcon` · `PersonPicture` · `EasingFunction`

`NormalizeMarkupSubstitutions` now resolves tokens by position:

| Position | Example | Handling |
|---|---|---|
| Attribute | `Click="X" $(IsEnabled)/>` | **Removed** — stands in for a whole attribute |
| Element content | `</AppBarButton>$(MoreCommands)` | **Becomes a comment** — all 5 upstream instances inject markup into a property element, never text, so flattening assigned the literal string `"..."` as a collection's content |
| Attribute value | `Value="$(Progress)"` | **Unchanged** — flattening here is cosmetic and stays valid |

Comments are copied verbatim so a `>` inside one cannot desynchronize the scanner.

## Usage Example

No CLI surface changes.

```console
$ winapp find-ui --id gallery-commandbar-1
```

**Before:** header, `Important` notes, `See also` — zero XAML, zero C#.

**After:** upstream's actual CommandBar markup, with the runtime-injection point marked as a comment.

## Related Issue

Unblocks [#703](https://github.com/microsoft/winappCli/issues/703). Companion to #806, which defines the index contract; this PR makes the data we would generate from that contract honest.

## Type of Change

- 🐛 Bug fix

## Checklist

- [x] New tests added for new functionality
- [ ] Main README.md updated — n/a
- [ ] docs/usage.md updated — n/a, no command surface changed
- [ ] Language-specific guides updated — n/a
- [ ] Sample projects updated — n/a
- [x] Shipped skills updated in `plugins/winapp/skills/` — `winapp-find-ui/SKILL.md`

### ⚠️ This PR re-bakes the corpus — do not apply `corpus-rebake-not-needed`

`snapshot-gallery.json.br`, `snapshot-reactor.json.br` and `snapshot-manifest.json` are updated, and **`CacheVersion` 19 → 21**.

The bump is load-bearing, not housekeeping. Caches match on an exact version string, so without it a user who already fetched keeps being served both the winapp-authored sample *and* the ten empty scenarios indefinitely — including offline. A re-bake alone never reaches them. `CacheVersion.cs` states the rule: bump whenever *"cleaning logic changes would alter the cached output for the same input data."*

### Isolating the change from upstream drift

The committed corpus was stale, so a naive before/after would have conflated our removal with upstream's own additions. Baked back to back against the same live upstream: `main` 323 scenarios, branch 322. The only differences were `itemsrepeater-7` removed and `tabview-1` restored to upstream's C# (939 → 1064 chars). The CommandBar injection produced no difference — direct confirmation it had gone dead.

Manifest now reads gallery 322 / reactor 95 / toolkit 48.

### Known follow-up (not in this PR)

`ItemsRepeater`'s `$(SampleCodeLayout)` resolves upstream to a `UniformGridLayout` block stored in a `Page.Resources` `x:String` that winapp does not fetch. Resolving those would recover more real upstream markup — and would have covered the image-grid case directly. Worth its own issue.

### Verification

- **4157** tests pass, 17 skipped. The single failure, `AnalyzeDumpAsync_ManagedDumpWithDeepStack_DetectsStackOverflowShape`, is an ARM64-host/x64-dump architecture mismatch **verified as pre-existing** by reproducing it on a stashed-clean tree. Network-bound classes (`Package`/`Sign`/`Nuget`/`Msix`/`BuildTools`/`EndToEnd`) are excluded — unreachable `api.nuget.org` on a corp machine; CI covers them.
- Three new tests cover the fetcher fix (attribute / value / content position). `GalleryCorpus_CarriesNoWinappAuthoredSample` still guards against an authored sample reappearing.
- `scripts/build-cli.ps1` clean end to end — **0 warnings**, warnings-as-errors on. `docs/cli-schema.json` unchanged, as expected for a no-surface-change PR.
- `scripts/validate-plugin-package.ps1` clean.
- Live CLI confirms all ten controls now serve XAML.


zateutsch:

## Decision: changes required

The core change — serve WinUI Gallery samples as upstream publishes them and drop the winapp-authored injections/overrides — is well-scoped, cleanly subtractive, and correctly compensated in most places. Two things to address before merge.

## Must fix

### `find-ui "commandbar"` advertises a code example but returns nothing
- **What is wrong:** Removing `InjectMissing` dropped the only CommandBar snippet. Upstream's CommandBar has no extractable ControlExample, so the sanitizer strips its code — but the control still appears with a fetchable id and the "Fetch full code" prompt.
- **Show me:** `winapp find-ui "commandbar"` → lists `gallery-commandbar-1` + "Fetch full code with: winapp find-ui --id <id>"; then `winapp find-ui --id gallery-commandbar-1` → exit 0 with only Important/Family/See-also notes, **zero XAML and zero C#**.
- **Why it matters:** Before this change the same query returned working, pasteable CommandBar code. Now a common control yields a successful-looking result an agent cannot copy — a straight regression to an empty result.
- **Smallest fix (consistent with this PR's thesis):** either suppress scenarios that extract to no code from the fetchable results, or move a CommandBar snippet into the attributed `Notes`/core-pattern mechanism (as done for the ItemsRepeater guidance) rather than back into the gallery corpus.
- **Location:** `src/winapp-CLI/WinApp.Cli/Services/Controls/GalleryFetcher.cs` (removed `InjectMissing`).

## Non-blocking

### "image grid" doesn't reach the new collection-layout guidance
- **What is wrong:** The compensating synonyms cover `photo`/`gallery`/`thumbnail`/`uniformgrid` but not the phrase "image grid" — which is literally the header of the deleted sample ("image grid (UniformGridLayout)").
- **Show me:** `winapp find-ui "image grid"` → only `Image` (nine-grid images, basic image); `winapp find-ui "photo grid"` → correctly routes to GridView/ItemsView.
- **Why it matters:** The most natural phrasing for the exact use case the new note was written for lands on single-image samples instead.
- **Smallest fix:** Add `image grid` / `image gallery` routing to `itemsrepeater`/`gridview`/`itemsview` in `Synonyms.cs`.
- **Location:** `src/winapp-CLI/WinApp.Cli/Services/Controls/Synonyms.cs`

### `gallery-tabview-1` now emits upstream C# that won't compile as-is
- **What is wrong:** Dropping `ApplyOverrides` means the served TabView sample keeps upstream's `frame.Navigate(typeof(YourPage))` with an undefined `YourPage`. That's upstream's own placeholder (deliberately deferred to now), but the new SKILL.md "compiles when pasted" phrasing oversells it.
- **Smallest fix:** Scope the compile claim to stripped handlers, or note that some samples reference page types the user must define.
- **Location:** `plugins/winapp/skills/winapp-find-ui/SKILL.md`

## What was exercised
- `dotnet build WinApp.Cli.csproj -c Release` — succeeded, 0 warnings / 0 errors.
- Ran the built binary directly: reproduced the empty-code CommandBar result and the "image grid" vs "photo grid" phrasing gap.
- Cache-version bump (19→20) and re-baked snapshot manifest verified consistent. No security-relevant path changed (fetch stays HTTPS from a Microsoft host).
- Not run independently: full MSTest suite. The new `GalleryCorpus_CarriesNoWinappAuthoredSample` test asserts one exact header is absent, so it guards the specific removed sample rather than the general invariant — acceptable but narrow.

zateutsch (changes requested):
Review attached above. Looks like there's a dead sample with the changes.

nmetulev (dismissed):
we should not have any authored content int he winapp cli - if content needs to exists, it needs to be pushed upstream to the gallery instead of adding it to winapp - let's use this pr to remove all winapp authored content and only leverage upstream content

plugins/winapp/skills/winapp-find-ui/SKILL.md
```diff
@@ -147,6 +147,30 @@ winapp find-ui "color picker" --json
 - Keep queries **focused** (one feature per query) — the lexical ranker rewards
   specific phrasing. Batch multiple focused queries rather than one broad one.
 
+## Upstream is the source of truth
```
nmetulev, line 150: do we need this in the skill? it sounds like internal mechanics instead of agent guidance

zateutsch (approved):
All clean, lgtm

PR #851, opened 2026-09-15
Add opt-in manual prerelease publish for library NuGet packages

zateutsch:
## What

Adds an opt-in, manual way to publish **prerelease** versions of the three source-built library NuGet packages to nuget.org, so they can be validated before the first official release:

- `Microsoft.Windows.SDK.BuildTools.WinApp.UIAutomation`
- `Microsoft.Windows.SDK.BuildTools.WinApp.UIAutomation.Recording`
- `Microsoft.Windows.SDK.BuildTools.WinUIAnalyzer`

## How

Queue **WinDevCLI - Release** manually with the new **`publishPrereleaseNugets`** parameter ticked (optionally set `prereleaseNugetVersion`; blank auto-computes `version.json` + `-prerelease.<build>`). This adds two stages — `Prerelease_Nugets_Build` and `Prerelease_Nugets_Publish` — that pack only those three packages (`package-nuget.ps1 -SkipCliPackage`, so the CLI tools package is never built), ESRP-sign them, and push only them via the `NuGet-WinAppCLI` service connection.

The two stages are **self-contained in `release.yml`** with inline ESRP signing; `build.yaml` is untouched.

## Why a parameter on the existing pipeline (not a new pipeline or branch)

- Reuses the release pipeline's service connections, variable groups, and 1ES onboarding — the only ADO wiring needed is allowing the run branch on the Branch control checks for `NuGet-WinAppCLI` + the signing connection.
- Stable library publishing is **unchanged** — it still ships through the normal `rel/v*` `Release_NuGet` stage.

## Safety model

The stages are gated at **compile time** so they can't misfire:

- `eq(parameters.publishPrereleaseNugets, true)` — defaults false, so the stages are pruned from the graph on every scheduled rehearsal and every real `rel/v*` release. They exist only when a human explicitly ticks the box.
- `eq(parameters.DoEsrp, true)` — the path is pruned entirely when signing is off, so it can **never publish unsigned packages**.
- `not(startsWith(Build.SourceBranch, 'refs/heads/rel/v'))` — belt-and-suspenders: even a manual queue on a release branch publishes nothing, so it can never ride a real release.

The `prereleaseNugetVersion` value is passed via an **environment variable** (not spliced into the inline script, avoiding PowerShell injection from a queue-time value) and validated against a strict SemVer-prerelease pattern, so a stable version — including stable-with-build-metadata like `0.7.0+build-1` — cannot be published through this path.

`Prerelease_Nugets_Build` also runs the targeted test suites for exactly these packages (`WinApp.UIAutomation.Tests` and the WinUI analyzer tests) before packing, so a test-failing commit can't be signed or published — while staying decoupled from the full ~60-min stable `Build`.

## Incidental fix: over-broad assertion in the winui-app sample test

`samples/winui-app/test.Tests.ps1:154` asserted `Should -Not -Match 'release-'` on the whole CLI output to check that the *release* publish profile wasn't selected. That scans the banner too, and on a branch build the banner prints a version that embeds the branch name (here `0.6.3-...-prerelease-pipeline.1`), which contains `release-` and false-fails the test. Tightened it to match the resolved `PublishProfile=release-` token specifically (mirroring the positive assertion on the line above), so it still catches a real release-profile regression without false-matching the version string. This is a latent bug that would fail any branch whose name contains `release-`; it is unrelated to the pipeline change but was surfacing as red CI on this branch.

## Review notes

Went through the repo's `pr-review` skill. It surfaced two HIGH findings (unsigned-publish reachable with `DoEsrp=false`; version-parameter injection + weak prerelease guard); both are fixed here and confirmed clean on re-review by an independent different-model cross-check. A reviewer also flagged that the decoupled build skipped tests — addressed with the targeted test gate above.

**Not yet validated:** an actual ADO template compile / preview and a real nuget.org push — `api.nuget.org` is unreachable from the corp dev environment. A dry compile in Azure DevOps on this branch is worth doing before first real use.

## Files

- `.pipelines/release.yml` — two parameters + the two self-contained prerelease stages (with the library test gate)
- `.pipelines/README.md` — "Publishing a prerelease of the library packages" section
- `samples/winui-app/test.Tests.ps1` — tighten the release-profile assertion (incidental CI fix)


azchohfi:
One thing worth flagging before this ships: the prerelease publish path bypasses the tested `Build` stage.

`Prerelease_Nugets_Build` is declared with `dependsOn: []`, and `Prerelease_Nugets_Publish` depends only on it — not on `Build`, which is where the test suite runs. So a manual prerelease queue will sign and push the three library packages to nuget.org even if the library tests would fail on that commit. Since a nuget.org push is immutable and Microsoft-signed, that means a test-catchable regression could go out publicly (only recoverable by bumping the prerelease number).

I get that decoupling is intentional to avoid waiting on the full ~60-min NativeAOT/MSIX build. If that tradeoff is deliberate, no change needed — just want to make sure it's a conscious choice. If not, the smallest fix is to add `Build` to the publish gate (`Prerelease_Nugets_Publish.dependsOn: [Prerelease_Nugets_Build, Build]`), or run just the library tests inside `Prerelease_Nugets_Build`.

zateutsch:
Good catch @azchohfi — this was a gap, not a conscious tradeoff, so I fixed it in 46214769.

You're right that decoupling from `Build` was intentional (to skip the ~60-min NativeAOT/MSIX build), but skipping the tests along with it wasn't the goal. Rather than gate on the full `Build` (whose suite also runs CLI tests irrelevant to these three packages), `Prerelease_Nugets_Build` now runs the targeted suites for exactly what ships — `WinApp.UIAutomation.Tests` (engine + recording) and the WinUI analyzer xUnit tests — before packing, mirroring `build-cli.ps1`'s runners. A failing suite fails the stage, so the dependent publish stage never runs. That keeps the path decoupled and fast while making sure a test-failing commit can't be signed and pushed to nuget.org.

azchohfi (approved):

PR #852, opened 2026-09-15
Never serve sample code that still contains a substitution placeholder

Jaylyn-Barbee:
`find-ui` promises code you can paste. For Gallery-style samples it could not keep that promise: those snippets carry `$(Name)` substitution placeholders that the Gallery resolves at runtime from live option controls, and nothing downstream was checking for them.

### The two ways this went wrong

**On the scraping path**, a placeholder in an attribute value was rewritten to the literal `"..."`. That keeps the markup well-formed, which is why it survived review as "cosmetic" — but it is not pasteable. Decompressing the previous bake and counting shows **140 attributes** set to `"..."`, and they are overwhelmingly *typed* properties:

| Property | Count | Why `"..."` fails |
| --- | --- | --- |
| `StrokeThickness`, `Height`, `Width`, `RadiusX/Y`, `Value` | 24 | not a `double` |
| `Orientation`, `SelectionMode`, `PaneDisplayMode` | 13 | not an enum member |
| `IsChecked` | 2 | not a `bool` |

Every one of those is a compile error the moment someone pastes it.

**On the index path**, nothing looked for placeholders at all, so a sample could be served with its tokens intact.

### What this changes

A shared detector, `SampleSubstitutionPlaceholder`, is now the single definition of what a placeholder is. Both paths use it.

**Scraping path.** `GalleryFetcher.NormalizeMarkupSubstitutions` resolves a token by the position it occupies rather than blanket-replacing it:

- *attribute-name position* (`<Button $(Background)/>`) — the whole attribute is removed
- *content position* — becomes a comment
- *value position* (`Value="$(X)"`) — **the whole attribute is dropped**, so the property falls back to its own default

The last one is the behaviour change. Dropping the attribute is what the Gallery-side exporter already does, and it is the only option that compiles.

Two supporting fixes were needed to make that actually take effect:

- `ExtractInlineCode` flattened XAML tokens to `"..."` *before* `CleanGalleryContent` ran, so the position-aware pass never saw them. It no longer pre-flattens.
- `CleanGalleryContent` now runs the known `IsOpen` / `Severity` rewrites *before* normalization, so those real values survive instead of having their attribute dropped.

**Index path.** `SampleIndexParser` suppresses a language block that still contains a placeholder and drops the sample only if neither block survives. `ScenarioSanitizer` enforces the same rule on the way out, so a placeholder cannot reach a caller through either route.

**Cache.** `CacheVersion` goes to `22`. Same input, different output, so without the bump an existing cache keeps serving the broken attributes this change exists to remove — the trap entry `"20"` already documents. The snapshot is re-baked accordingly.

### Result

Measured against a fresh bake, decompressing each snapshot and parsing the JSON:

| Source | Scenarios | XAML | Attributes set to `"..."` | Raw tokens |
| --- | --- | --- | --- | --- |
| gallery | 330 | 310 | 0 (was 140) | 0 |
| toolkit | 48 | 48 | 0 | 0 |
| reactor | 95 | 0 | 0 | 0 |

No content was lost making this safe: gallery XAML went **up**, 302 → 310, because fragments that previously failed the sanitizer's well-formedness check now survive.

### Schema

The contract gains the three `gallery` fields the WinUI Gallery exporter actually emits, which were undocumented:

- `xamlPlaceholdersDropped` — tokens already removed; the XAML pastes as published
- `xamlOmittedAsMalformed` — XAML withheld because it did not parse (12 samples upstream)
- `codePlaceholdersPresent` — tokens **still present** in `code`; it does not compile as published

Read the first and last as opposites, not companions. One records cleanup already done, the other warns that cleanup was not possible. A consumer conflating them pastes broken C#, which is what the field exists to prevent — the description says so explicitly.

C# gets no deletion equivalent, and that is a property of the language rather than unfinished work: no C# construct yields a default by being absent. The tokens that actually occur are identifier fragments (`BadgeNotificationGlyph.$(SelectedGlyph)`), fixed-arity arguments (`SetBorderAndTitleBar($(HasBorder), $(HasTitleBar))`), and whole statements. None can be cut out without a syntax error or removing the sample's subject.

### Testing

Full suite: **5437 tests, 12 failures**, all pre-existing and environmental — 11 NuGet live-source tests that need `api.nuget.org`, and 1 crash-dump test. Same 12 fail on a clean checkout.

Directly affected classes re-run green after the schema change: `SampleIndexTests` 25/25 (includes the schema/reader drift guard), `GalleryFetcherCleanContentTests` 7/7, `FindUiSearchTests` 27/27, `ScenarioSanitizerTests` 23/23, `EmbeddedSnapshotTests` 19/19.

New tests cover each token position, the attribute-drop behaviour, and a guard that the `IsOpen`/`Severity` rewrites still win over attribute dropping — the regression the reordering could have caused.

### Related

Related to microsoft/WinUI-Gallery#2232, which publishes the sample index and guarantees its XAML is placeholder-free. The two are **technically independent and can merge in either order** — the Gallery's index validates against this repo's schema both before and after this PR, because the contract sets `additionalProperties: true` at every level. Confirmed with ajv (draft 2020-12) against both the current `main` schema and the one here: valid in both cases.

What this PR adds is description rather than permission. On `main`, `sample.gallery` declares no properties at all, so the three fields that carry the pasteability contract are tolerated but unexplained. Landing this first simply means `codePlaceholdersPresent` is described by the published schema as soon as the Gallery starts emitting it, rather than shortly after.


zateutsch (approved):
Approving with one non blocking item from AI review, we should probably remove this unless we have a reason for having it:

### Published `gallery` schema object has no producer or consumer

- **What is wrong:** The PR adds a `sample.gallery` object (`xamlPlaceholdersDropped`, `xamlOmittedAsMalformed`, `codePlaceholdersPresent`) to the published sample-index schema, but nothing writes those fields and nothing reads them. It documents a diagnostic contract that doesn't exist yet.

- **Show me:** `git grep xamlPlaceholdersDropped` → only `docs/winui-sample-index.schema.json`. `SampleIndexSchema.Gallery` is referenced only in `SampleProperties`, which is consumed only by a test parity assertion (`SampleIndexTests.cs:500`). No producer emits it; the parser suppresses placeholder blocks without reading it.

- **Why it matters:** A published schema is effectively forever. An external producer may implement these fields believing they're honored, and future maintainers must preserve or explain a contract the code never implemented.

- **Smallest fix:** Drop the `gallery` object from `docs/winui-sample-index.schema.json` and remove `Gallery` from `SampleProperties` until a real writer/reader needs it. Keep the pasteability wording changes to `xaml`/`code`.

- **Location:** `docs/winui-sample-index.schema.json:148`, `src/winapp-CLI/WinApp.Cli/Services/Controls/SampleIndexSchema.cs:69,86`

PR #878, opened 2026-09-18
Fix empty UI values falling through to accessibility labels

### LINKED ISSUE #875: [Bug]: ui get-value returns a text box's label when its value is empty
### Describe the bug

`winapp ui get-value` can return an empty text box's accessibility label instead of its actual empty value. For example, an empty title field is reported as containing `Intention title, required`.

This is a **shared UI Automation value-reading bug, not a sandbox-specific bug**. It was originally observed in Windows Sandbox, but independently reproduced against the shared implementation using an isolated test with simulated UI Automation providers. A paired live local/sandbox reproduction has not been performed.

Agents and tests cannot reliably determine whether a field is empty. `wait-for --value ""` also uses the same value-reading method, so it can fail to match an existing empty field.

### To Reproduce

**Independent, isolated reproduction** (source revision `2e307c7e8e648a1906205241148de20bcf12f7ab`):

```powershell
dotnet run --project src\winapp-CLI\WinApp.UIAutomation.Tests\WinApp.UIAutomation.Tests.csproj -c Debug -- --filter "FullyQualifiedName~FaultInjectedComProxies_CoverEmptyTextAndValueFallthrough"
```

The test supplies successful empty TextPattern and ValuePattern reads and an accessibility name of `Fallback Name`. The actual shared `GetTextAsync` implementation returns `Fallback Name`, not an empty string. The test passed on September 18, 2026 because it currently **asserts the incorrect fallback behavior**; that pass confirms the defect rather than validating the desired behavior.

Relevant source at the tested revision:
- [GetTextAsync](https://github.com/microsoft/winappCli/blob/2e307c7e8e648a1906205241148de20bcf12f7ab/src/winapp-CLI/WinApp.UIAutomation/Services/UiAutomationService.cs#L963-L1052): skips successful empty TextPattern and ValuePattern reads before falling back to Name.
- [Existing isolated test](https://github.com/microsoft/winappCli/blob/2e307c7e8e648a1906205241148de20bcf12f7ab/src/winapp-CLI/WinApp.UIAutomation.Tests/RealUiAutomationTests.Coverage.cs#L1364-L1407): expects `Fallback Name` for an empty control.

**Originally reported UI example** (not independently rerun against the live app): with an empty labeled text box selected by `TaskTitleInput`, `get-property ... -p Value --json` returned `{"properties":{"Value":""}}`, while `get-value ... --json` returned `{"text":"Intention title, required"}`. The report used CLI `0.6.3-prerelease.52` in an ARM64 sandbox.

### Expected behavior

A successfully read empty text/value is valid and should be returned as `""`. An accessibility label must not replace it simply because it is empty.

Recommended fix:
- Preserve successful empty reads from supported text/value patterns; retain the existing intentional pattern precedence.
- Fall back to the accessibility name only when the appropriate text/value source is unavailable, not when it returns an empty string.
- Correct the existing test and add regression coverage for empty, whitespace-only, and nonempty values, plus fallback when patterns are unsupported.
- Verify that `get-value` and `wait-for --value ""` agree for a fresh empty field and a field cleared after editing.

Do not remove the app's accessibility label or weaken assertions as a workaround.

### Screenshots

N/A. The independent reproduction uses simulated UI Automation providers and requires no app window.

### OS Version and details

Independent isolated reproduction: Windows ARM64, .NET 10.0.12, MSTest 4.1.0, repository revision `2e307c7e8e648a1906205241148de20bcf12f7ab`.

Original live report: CLI `0.6.3-prerelease.52`, native ARM64 host and Windows Sandbox. The bug was independently confirmed in shared code; live host-versus-sandbox comparison remains untested.

nmetulev:
## Description

Empty text fields return their actual empty value instead of their accessibility label. Successful TextPattern and ValuePattern reads preserve empty, whitespace-only, and nonempty text while retaining the existing pattern precedence, unsupported-pattern fallback, exception handling, and result shapes. Both allocated-empty and null BSTR representations are treated as empty strings, using the existing ownership-aware BSTR reader.

## Usage Example

Illustrative commands for an empty field with the accessibility label `Title, required`:

```powershell
winapp ui get-value TaskTitleInput -a myapp --json
# Before: "text": "Title, required"
# After:  "text": ""
winapp ui wait-for TaskTitleInput -a myapp --value "" --timeout 5000
# Matches a fresh empty field or a field cleared after editing.
winapp ui get-property TaskTitleInput -a myapp --property Name
# Reads the unchanged accessibility label separately.
```

## Related Issue

Closes #875.

## Type of Change

- Bug fix
- Documentation
- Test update

## Checklist

- [x] Added regressions that fail before the fixes
- [x] Tested locally on Windows ARM64 with simulated providers and command fakes
- [x] Updated the UI automation guide and shipped skill
- [x] Built native CLI/packages and validated generated docs

## Screenshots / Demo

N/A: this is a value-reading change, demonstrated with deterministic provider and command tests rather than visual changes.

## Additional Notes

Focused validation on the updated source:

```powershell
dotnet run --project src\winapp-CLI\WinApp.UIAutomation.Tests\WinApp.UIAutomation.Tests.csproj -c Debug -- --filter "FullyQualifiedName~FaultInjectedComProxies"
# 18 passed, including allocated-empty and null-BSTR reads for both patterns.
dotnet run --project src\winapp-CLI\WinApp.Cli.Tests\WinApp.Cli.Tests.csproj -c Debug -- --filter "FullyQualifiedName~UiCommandTests.GetValue|FullyQualifiedName~UiCommandTests.WaitFor"
# 33 passed, including exact JSON content and fresh/cleared empty waits.
.\scripts\build-cli.ps1 -SkipTests -SkipMsix
# Passed: x64/ARM64 NativeAOT, npm and four NuGet packages, generated docs.
.\scripts\validate-llm-docs.ps1 -CliPath .\artifacts\cli\win-arm64\winapp.exe
# Passed: schema drift and plugin conformance.
```

The allocated-empty regressions reproduced three failures before the initial fix; the null-BSTR regressions reproduced two failures before normalization. Current CI and review gates are tracked in the [lifecycle status comment](https://github.com/microsoft/winappCli/pull/878#issuecomment-5766957812); prior-head green checks are not current readiness evidence.

No local host UI or shared Windows Sandbox was touched. Full/live tests run in repository CI; a paired host/sandbox comparison has not been performed. Focus reporting and activation changes are out of scope.


src/winapp-CLI/WinApp.UIAutomation/Services/UiAutomationService.cs
```diff
@@ -1085,10 +1085,7 @@ public Task FocusAsync(UiTarget uiTarget, UiElement element, CancellationToken c
             patternAcquired = true;
             var range = pattern.get_DocumentRange();
             var text = range.GetText(-1);
-            if (text.Length > 0)
-            {
-                return Task.FromResult<string?>(text.ToString());
-            }
+            return Task.FromResult<string?>(text.ToString());
```
zateutsch: This return and the `ValuePattern` return just below (line 1100) both call `.ToString()` on the UIA BSTR directly. When a provider represents an empty value as a null (`default`) BSTR, `BSTR.ToString()` returns `null` — so this returns `null` instead of `""`, and `wait-for --value ""` times out while `get-value --json` shows `"text": null`.

Every other BSTR read in this class routes through `GetBstr(...)`, which coalesces null → `""` *and* frees the UIA-owned BSTR (the direct `.ToString()` here also leaks it). Suggested:

```csharp
return Task.FromResult<string?>(GetBstr(() => range.GetText(-1)));          // line 1088
return Task.FromResult<string?>(GetBstr(() => pattern.get_CurrentValue())); // line 1100
```

`GetBstr` still lets getter exceptions propagate to the existing `catch` filter. Worth adding a `default(BSTR)`-on-a-supported-pattern row to `FaultInjectedComProxies_GetTextPreservesSuccessfulReads` to lock it in.

zateutsch (changes requested):

**Request changes** — the fix is clean and well-tested for the common case, but it only handles the *allocated* empty-string representation. When a UIA provider returns an empty field as a null (`default`) BSTR — which this repo already treats as `""` elsewhere (see `Query_GeneralStringNullAndEmptyBstrRemainEmpty`) — the new direct `.ToString()` returns `null`. So `winapp ui get-value --json` emits `"text": null` and `winapp ui wait-for --value ""` never matches, which defeats the documented fix for those providers. See the inline comment for the smallest fix.

Verified locally: the solution builds with zero warnings and the changed tests pass (8 UI Automation rows + 5 CLI). Precedence, the static-label → `Name` fallback, and the docs/skill updates all look correct.

nmetulev:
**State:** merged — lifecycle monitoring stopped.
**Assessed:** GitHub reports PR merged; final PR head `bf291b7789463b0dea20c2c56458b0855ca76fa6`.
**Evidence:** [final-head Build and Package](https://github.com/microsoft/winappCli/actions/runs/35661764990) succeeded; required build-and-package, validation lanes, UI E2E, docs and samples passed. Reusable sample build intentionally skipped.
**Review:** zateutsch submitted APPROVED on the final head, superseding the earlier changes request. Both actionable threads were answered and resolved. The agent did not dismiss a review or perform the merge.
**Pending:** none for this PR lifecycle.
**Resume:** stopped; active lifecycle labels removed and same-session automation cleared. No branch recreation or further code changes.

src/winapp-CLI/WinApp.UIAutomation/Services/UiAutomationService.cs
nmetulev: Fixed in f69c0b6010bcc1d4e0138f947d6c5dc4dcde97b5. Both successful pattern reads now use the existing GetBstr helper, which normalizes a null BSTR to `""` and frees transferred BSTR memory while leaving getter failures to the existing catch filters. Added default(BSTR) cases for TextPattern (with a lower-priority nonempty ValuePattern) and ValuePattern; both failed before the fix and now pass. All 18 fault-injected provider cases and 33 get-value/wait-for command cases pass locally. Unsupported-pattern fallback, whitespace/nonempty values, labels and precedence remain covered. Latest main was merged without rewriting history. Re-requested your review; the changes-request gate remains pending your reassessment, not dismissed.

plugins/winapp/skills/winapp-ui-automation/SKILL.md
```diff
@@ -320,6 +320,10 @@ winapp ui get-property Document -a myapp --property FontWeight --json
 winapp ui get-focused -a myapp
 ```
 
+To check a blank or cleared field, use `wait-for SearchBox -a myapp --value ""`.
```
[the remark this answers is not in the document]
nmetulev: Fixed in 9f1aaf0c (included in current head 01d19bc8): the example now reads `winapp ui wait-for SearchBox -a myapp --value ""`. Validated plugin conformance and the native CLI's `winapp ui wait-for --help` usage. No command behavior changed.

zateutsch (approved):

